PR #13, opened 2019-05-03
Clarify relationship with other proposals

littledan:


rricard:
It is indeed more clear, thank you!

PR #16, opened 2019-06-06
Minor typo. map1 -> map2

rictic:
(assuming I'm following along correctly)


rricard:
Thank you for keeping an eye on this! #14 has fixed it though and I just merged the first one on the queue!

PR #22, opened 2019-06-11
Added a section to clarify usage inside standard collections Map/Set/WeakMap

rickbutton:
I've added a small section that attempts to describe the interaction with `Map`, `Set`, and `WeakMap`, as well as an example of each.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,7 +218,7 @@ assert("a" in @const { a: 1, b: 2 });
 
 It is possible to use a const object or const array as a key in a `Map`, and as a value in a `Set`. When using a const object or const array in this way, key/value equality behaves as expected.
 
-It is not possible to use a const arary of object as a key in a `WeakMap`, because const arrays and const objects are not `Objects`, and have no lifetime. Attempting to set a value in a WeakMap using a const object
+It is not possible to use a const object or const array as a key in a `WeakMap`, because const objects and const arrays are not `Objects`, and have no lifetime. Attempting to set a value in a WeakMap using a const object
 or const array as the key will result in a `TypeError`.
 
 ### Examples
--- a/README.md
+++ b/README.md
@@ -214,6 +214,48 @@ assert(Object.keys(@const { a: 1, b: 2 }) === @const ["a", "b"]);
 assert("a" in @const { a: 1, b: 2 });
 ```
 
+## Usage in {`Map`|`Set`|`WeakMap`}
+
+It is possible to use a const object or const array as a key in a `Map`, and as a value in a `Set`. When using a const object or const array in this way, key/value equality behaves as expected.
+
+It is not possible to use a const arary of object as a key in a `WeakMap`, because const arrays and const objects are not `Objects`, and have no lifetime. Attempting to set a value in a WeakMap using a const object
+or const array as the key will result in a `TypeError`.
+
+### Examples
+
+#### Map
+
+```js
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -214,6 +214,48 @@ assert(Object.keys(@const { a: 1, b: 2 }) === @const ["a", "b"]);
 assert("a" in @const { a: 1, b: 2 });
 ```
 
+## Usage in {`Map`|`Set`|`WeakMap`}
+
+It is possible to use a const object or const array as a key in a `Map`, and as a value in a `Set`. When using a const object or const array in this way, key/value equality behaves as expected.
+
+It is not possible to use a const object or const array as a key in a `WeakMap`, because const objects and const arrays are not `Objects`, and have no lifetime. Attempting to set a value in a WeakMap using a const object
```
ljharb: Why would they have no lifetime? If i make a massive const object, and then it becomes collectible, why would we prohibit engines from doing so?

ljharb: More to the point, an object is any value for which `Object(x) === x`, so these are definitely objects.

rricard: It is collectible by the engine but like a string: the lifetime of the const object will not be observable.

ljharb: why deviate from language precedent? it's an object.

README.md
```diff
@@ -221,6 +221,48 @@ The const array prototype is a const object that contains the same methods as Ar
 - `ConstArray.prototype.pop()` and `ConstArray.prototype.shift()` do not return the removed element, they return the result of the change
 - `ConstArray.prototype.first()` and `ConstArray.prototype.last()` are added to return the first and last element of the const array
 
+## Usage in {`Map`|`Set`|`WeakMap`}
+
+It is possible to use a const object or const array as a key in a `Map`, and as a value in a `Set`. When using a const object or const array in this way, key/value equality behaves as expected.
+
+It is not possible to use a const object or const array as a key in a `WeakMap`, because const objects and const arrays are not `Objects`, and their lifetime is not observable. Attempting to set a value in a WeakMap using a const object
```
ljharb: this is something i'll bring up in committee when the time arises; these are certainly objects if `Object(x) === x`, and `typeof x` is `'object'` or `'function'`.

rricard: `Object(x) !== x`, because `Object(x)` returns a mutable object.

`typeof x` will be `constobject` or `record` or whatever thing we need to decide on (i think that @rickbutton is working on an another PR to clarify that)

rricard: And I did not check but #23 is that PR!

ljharb: no, it does not return a mutable object - `Object(Object.freeze(x))` for Object x returns a frozen object.

rickbutton: exactly, `Object(x)` will copy everything into a new mutable/regular/`typeof z === "object"` object.

ljharb: in that case, it sounds like this proposal defines new primitives, so “const object” would be a misleading name.

rricard: We do not have the same semantics as `Object.freeze` here. Yes we'd copy everything to the new mutable object

rricard: > in that case, it sounds like this proposal defines new primitives, so “const object” would be a misleading name (i see the proposal is named const value types, so i must have gotten confused seeing that verbiage somewhere)

Yes we're kind of trying to find a new name instead of const value types. I think I'll open an issue soon about it

pipobscure:
LGTM

PR #23, opened 2019-06-11
Added section to specify that typeof will return a new value

rickbutton:
The actual value hasn't been decided on, but I added a section that describes that the `typeof` const objects/arrays is not `object`, but some other value (placeholder for now).

This will conflict with my other PRs, I will take care of rebasing etc when the time comes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,16 @@ The const array prototype is a const object that contains the same methods as Ar
 - `ConstArray.prototype.pop()` and `ConstArray.prototype.shift()` do not return the removed element, they return the result of the change
 - `ConstArray.prototype.first()` and `ConstArray.prototype.last()` are added to return the first and last element of the const array
 
+## `typeof`
+
+The typeof operator will return a new value for const objects and const arrays. The value to be returned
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -221,6 +221,17 @@ The const array prototype is a const object that contains the same methods as Ar
...
 
+
+## `typeof`
+
+The typeof operator will return a new value for const objects and const arrays. The value to be returned
+is still being considered, and is represented by `<placeholder>` below.
+
+```js
+assert(typeof @const { a: 1 } === "<placeholder>");
+assert(typeof @const [1, 2]   === "<placeholder>");
```
ljharb: if it's not an object (or a function), then that suggests it's a primitive. which means `Object(x) === x` needs to be false (and `Object(x)` needs to created a boxed object form of the primitive). Is that the plan?

rricard: As explained in #22: `Object(x) !== x`, I think this is necessary otherwise, this will introduce a huge performance cliff in the engine if we implement structural sharing for const objects

PR #28, opened 2019-06-11
Clarify that const arrays are iterable, while const objects are not

rickbutton:
Just like their mutable counterparts, `const objects` are not iterable, while `const arrays` are.


rricard:
Fixes #28

PR #32, opened 2019-06-26
Added new section in the FAQ about # vs keyword, cleaned up styles

rickbutton:
I added a new section in the FAQ about `# vs keyword` pros/cons.

I also cleaned up a bit of the style. The doc is getting a bit long, so bumping up the section headers one size, and the subheaders one size, helps with readability IMO.

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,6 +343,17 @@ a similar concept (variable reference immutability) while this proposal intends 
 While immutability is the common thread between these two features, there has been significant community feedback that
 indicates that using `const` in both contexts is undesirable.
 
+Instead of using a keyword, `{| |}` and `[||]` have been suggested as possible alternatives. For example:
+
+```js
+const first = {| a: 1, b: 2 |};
+const second = [|1, 2, 3|];
+```
+
+This syntax also avoids the problems with using a keyword. However, it is also used by [Flow](https://flow.org/) as
+the syntax for [exact object types](https://flow.org/en/docs/types/objects/#toc-exact-object-types). Investigation
+will need to be done to determine if introducing this syntax in ECMAScript will break existing Flow typings.
+
 ## How does this relate to the const keyword?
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -326,15 +326,30 @@ const weakMap = new WeakMap();
...
+that match the "intent" of the feature, other than `const`. The `const` keyword is also tricky, because it describes
+a similar concept (variable reference immutability) while this proposal intends to add new immutable data structures.
+While immutability is the common thread between these two features, the features are orthogonal, and using `const`
+in both contexts would create hard-to-reason-about ambiguities.
 
-### Relation to const?
+Using the `#` character solves most of these problems. The `#` character can be safely used in this context without
+re-using an existing keyword or reserving a new keyword. Further, while `#` is already used to describe
+private fields in the [Class field declarations](https://github.com/tc39/proposal-class-fields) proposal, it is
+less ambiguous than re-using `const`, because they are completely orthogonal features (and contexts) with no relation
```
ljharb: This is a very subjective claim; i find it more ambiguous *because* they’re not related features.

rricard:
Can you also propose the `{| |}`/`[| |]` alternative in that writeup as well as linking to a list of the reserved keywords.

rickbutton:
@rricard can do!

README.md
rricard: I agree, maybe saying that `const` had a strong community pushback would be more accurate

rickbutton: I agree, probably a little too much personal opinion, not enough fact. I've reworded it to better fit @rricard's suggestion of referencing the community pushback against `const` alone.

rickbutton:
@rricard added the `{||}` and `[||]` blurb as requested, and added link to a reserved keyword list.

rickbutton:
thanks @ljharb for the always super fast reviews!

Issue #24, opened 2019-06-11
What is a const object?


rricard:
So we're going to dump the name `const object` and favor a more different naming (for the proposal and the results of a `typeof`)

For now we have multiple ideas:

- Immutable objects/arrays - `typeof immObjOrArr === "immutable"`
- Records/Tuples - `typeof recordOrTuple=== "record"`
- _Your idea here!_

@ljharb, I am very interested in what you think!

ljharb:
While I'm sure I'm missing some, and it's entirely possible some of these are incorrect or do not hold as absolutely as i'm implying, here's some current axioms I observe:

Objects:
 - can be used as WeakSet values, WeakMap keys, and a WeakRef target
 - `Object(x) === x` is true
 - `typeof x` is `'object'` or `'function'` (this list, of course, could grow if needed, but there's a web compat risk)
 - can have own properties; and can be frozen/sealed/preventExtensioned.

Primitives:
 - can not be used in weak contexts
 - have no properties - *all property access whatsoever* is done by creating an ephemeral boxed object, and accessing values on the prototype. `Object.keys(x)` and `Object.getOwnPropertyNames(x)` of any non-nullish primitive is always an empty array.
 - `Object(x)` creates a new boxed primitive, such that `Object(x) !== x` for any primitive `x`
 - `typeof x` is `'boolean'` or `'undefined'` or `'string'` or `'number'` or `'symbol'`, or `'bigint'`, or `'object'` for `null` (this list, of course, could grow if needed, but there's a web compat risk - adding symbols broke some code on the web, which caused it to use alternative means mentioned elsewhere in this comment)

Non-nullish instances, whether objects or primitives:
 - have a global Constructor, and a `.prototype` on that constructor
 - have at least one `Constructor.prototype.something.call(x)` method does an internal slot check on `x`, typically throwing when `x` lacks the slot (`Error` is the lone legacy exception; `Array.isArray` is a static predicate instead of an instance thrower)

It seems like this proposal seeks to violate some of these axioms.

In particular, note that primitives *do not have any own properties* - only properties shared via a prototype. If it has any own properties, *it is an object*, and must conform to the precedents set above.

So, what kind of values are produced by this proposal, taking into account the axioms

JustFly1984:
My 2 cents: What about immutable functions?

littledan:
@ljharb The idea of this proposal is that records and tuples are primitives. What do you see violated?

littledan:
[...]

PR #34, opened 2019-07-23
[WIP] Initial pass of converting to Record/Tuple naming

rickbutton:
[WIP]! I likely missed something / broke something, so this needs another pass.

This is the first pass of renaming "const values types" -> "Record" and "Tuple". I attempted to replace all usage/reference to the new types as objects/constant value types with their proper names. This required slightly changing the wording in some places in order to be less ambiguous. 

So far, I have left the `Syntax` section unchanged, and they still refer `const`. What do we want to do here?

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,6 +432,6 @@ Structural sharing is a technique used to limit the memory footprint of immutabl
 
 #### Value type
 
-In this proposal it defines any of those: `boolean`, `number`, `symbol`, `string`, `undefined`, `null`, `record` and `tuple`.
+In this proposal it defines any of those: `boolean`, `number`, `symbol`, `string`, `undefined`, `null`, `Record` and `Tuple`.
 
 Value types can only contain other value types: because of that, two value types with the same contents are strictly equal.
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# Constant Value Types
+# Record/Tuple Value Types
 
-ECMAScript proposal for constant and value types (also known as immutable types).
+ECMAScript proposal for the Record and Tuple value types (also known as immutable types).
 
 **Authors:**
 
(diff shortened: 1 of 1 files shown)


kleinfreund:
I find these names rather confusing.

Neither tuple nor record convey that their contents are immutable. The phrase “constant value types” however does focus on specifically that: A type whose values are constant (i.e. not subject to change). Python makes the arbitrary distinction of lists being mutable while tuples are not. That’s hardly enough precedent to call it common knowledge. Calling immutable objects records would be just as arbitrary. Record is synonymous to what an object is in JavaScript: An unordered collection of fields of potentially different types.

I coulnd’t find any discussion around why these names were chosen.

rickbutton:
I believe record/tuple as an option was mentioned in this issue: https://github.com/rricard/proposal-const-value-types/issues/24 , but not much discussion re: naming has happened in that issue so far.

One of the reasons that I believe `Record` and `Tuple` might be a better choice is that there was a decent amount of confusion in the feedback about whether these new value types were "objects" or not. The current summary on `master` references these values as `const object` and `const array`, but that is not accurate because they are not objects or arrays(which are objects) and do not behave like objects or arrays. Calling them `const object` and `const array` implies (to me, at least) that they are simply objects with addition of immutability (like using Object.freeze), but that is not the intent of the proposal, as these are completely independent types.

kleinfreund:
Is there a sentence somewhere indicating what the intention of this proposal is if it’s not adding “simply objects with addition of immutability”? Because that’s certainly how I understood the intention of the proposal to be: Objects (Arrays), but you can’t mutate them.

What is it that makes the entities of this proposal fundamentally different from having them be called immutable objects?

rickbutton:
The aforementioned issue, #24, discusses the fact that they are intended to be primitives, which implies that they are not objects.

[This comment](https://github.com/rricard/proposal-const-value-types/issues/24#issuecomment-501029962) provides a fantastic example of (some, but not all) of the differences between an object and a primitive.

PR #35, opened 2019-07-31
Fixes typo in syntax definitions

bkrem:
* Corrects `ConstUpdateExpresion` to `ConstUpdateExpression`


rickbutton:
thanks! good catch

PR #42, opened 2019-08-17
Add JSON roundtripping goal and consequences

### LINKED ISSUE #39: Interactions with Symbols
Assuming we want const values to allow efficient cloning across process boundaries, we need to define serialization.  How will Symbols be handled in this case?

mheiber:
- Add Object.from(record)
- Add behavior of JSON.stringify(tuple) and
JSON.stringify(record)
- Clarify what values are allowed as
keys/values of records and as entries in tuples

fix #39

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,19 +22,20 @@ ECMAScript proposal for the Record and Tuple const value types (also known as im
 
 # Overview
 
-The goal of this proposal is to introduce deeply constant/immutable value types to JavaScript. It has multiple objectives:
+The goal of this proposal is to introduce deeply constant/immutable data structures to JavaScript. It has multiple objectives:
 
 - Introducing efficient data structures that makes copying and changing them cheap and will allow programs avoiding mutation of data to run faster (pattern heavily used in Redux for instance).
 - Add guarantees in strict equality when comparing data. This is only possible because those data structures are deeply immutable (comparing props fast is essential for efficient virtual dom reconciliation in React apps for instance)
 - Be easily understood by external typesystem supersets such as TypeScript or Flow.
 - Offers the possibility to improve structured cloning efficiency when messaging across workers.
+- Supports lossless round-tripping with JSON.
 
 This proposal presents 2 main additions to the language:
 
 - `Record`
 - `Tuple`
 
-The only valid sub-structures of these values will be one of those structures and normal value types such as `number`, `string`, `symbol` or `null`.
+`Record`s and `Tuple`s can only contain other immutable values.
 
 ## Prior work on immutable data structures in JavaScript
 
@@ -137,6 +138,7 @@ Array.from(tuple).map(x => new MyClass(x))
 assert((#{} with .a = 1, .b = 2) === #{ a: 1, b: 2 });
 assert((#[ {} ] with [0].a = 1) === #[ { a: 1 } ]);
 assert((x = 0, #[ {} ] with [x].a = 1) === #[ { a: 1 } ]);
+assert.equal(#{ a: 1 }, Record.from(JSON.parse(JSON.stringify(#{ a: 1 }))));
 ```
 
 # Syntax
@@ -167,6 +169,12 @@ _ConstExpression_:
 #### Runtime verification
 
 At runtime, if a non-value type is placed inside a `Record` or `Tuple`, it is a `TypeError`. This means that a `Record` or `Tuple` expression can only contain value types.
+
+At runtime, attempting to create a `Record` with a key that is not a `string` is a `TypeError`.
+
+At runtime, it is a `TypeError` to add a value to a `Record` or `Tuple` of any type except the following: `Record`, `Tuple`, `string`, `number`, `true`, `false`, and `null`.
+
+> `undefined` is not permitted, as that would violate the goal of lossless round-tripping with JSON.
 
 ## Const update expression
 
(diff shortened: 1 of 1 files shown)


ljharb:
That also precludes including frozen functions - then the proposal is not really about immutability, it’s about json round tripping, which means that mutable objects should be allowed if they can json round trip.

ljharb:
What does Object.from do with things that aren’t records or tuples?

mheiber:
@ljharb I'm doubting whether Object.from was a good idea, since there is already so much weird stuff in JS. What about co-opting `new Object(record)` and `Object(record)` instead? They already use the [`toObject`](https://www.ecma-international.org/ecma-262/10.0/index.html#sec-toobject) abstract op that says how to make an object from each kind of thing.

ljharb:
`Object(record)` must make a boxed primitive object that's `instanceof Record`; if you want to convert it to a plain object, you need something different.

mheiber:
@ljharb can you use line comments? Unfortunately GH seems to only make threads if a comment is related to a line. 
You [said](https://github.com/rricard/proposal-const-value-types/pull/42#issuecomment-522264893)

> That also precludes including frozen functions - then the proposal is not really about immutability, it’s about json round tripping, which means that mutable objects should be allowed if they can json round trip.

1. What alternative are you suggesting?

2. Is the issue that you see immutability and simple (de)serialization as non-orthogonal goals? If so, note that the proposal is already not 100% orthogonal:
- We could add immutable objects where `===` is pointer equality. In fact, JS libraries offer those already! Examples include ImmutableJS, Mori, and MicroState.
- We could add mutable objects where `===` is structural. In fact, we could do that very easily by having a new well-known symbol, `Symbol@@equals`, thereby allowing userland implementations of arbitrary equality-checking algorithms.

There are tradeoffs here. My hope is to bias toward providing **good performance and developer ergonomics** at the expense of some orthogonality.

3. Would you agree that it will be beneficial to have predictable and useful serialization?

I believe the proposal was designed with ease of serialization/deserialization in mind, such as communicating data with workers. Keeping serialisation simple for now helps us build toward that goal.

You specifically mentioned functions:
    - You mentioned frozen functions specifically. There is no special-casing for freezing in the current proposal. `#{ a: Object.freeze({}) }` errors if I understand correctly.
    -  Functions are particularly nasty if we want a good (de)serialisation story. They aren't supported by JSON.stringify by default, and aren't supported at all by the [HTML5 structured clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm).

README.md
```diff
@@ -22,19 +22,20 @@ ECMAScript proposal for the Record and Tuple const value types (also known as im
 
 # Overview
 
-The goal of this proposal is to introduce deeply constant/immutable value types to JavaScript. It has multiple objectives:
+The goal of this proposal is to introduce deeply constant/immutable data structures to JavaScript. It has multiple objectives:
```
mheiber: @ljharb, you [said](https://github.com/rricard/proposal-const-value-types/pull/42#issuecomment-522264893)

> That also precludes including frozen functions - then the proposal is not really about immutability, it’s about json round tripping, which means that mutable objects should be allowed if they can json round trip.

1. What alternative are you suggesting?

2. Is the issue that you see immutability and simple (de)serialization as non-orthogonal goals? If so, note that the proposal is already not 100% orthogonal:
- We could add immutable objects where `===` is pointer equality. In fact, JS libraries offer those already! Examples include ImmutableJS, Mori, and MicroState.
- We could add mutable objects where `===` is structural. In fact, we could do that very easily by having a new well-known symbol, `Symbol@@equals`, thereby allowing userland implementations of arbitrary equality-checking algorithms.

There are tradeoffs here. My hope is to bias toward providing **good performance and developer ergonomics** at the expense of some orthogonality.

3. Would you agree that it will be beneficial to have predictable and useful serialization?

I believe the proposal was designed with ease of serialization/deserialization in mind, such as communicating data with workers. Keeping serialisation simple for now helps us build toward that goal.

You specifically mentioned functions:
    - You mentioned frozen functions specifically. There is no special-casing for freezing in the current proposal. `#{ a: Object.freeze({}) }` errors if I understand correctly.
    -  Functions are particularly nasty if we want a good (de)serialisation story. They aren't supported by JSON.stringify by default, and aren't supported at all by the [HTML5 structured clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm).

ljharb: I guess I'm a bit confused. If this proposal is about immutability, then any immutable object - including deeply frozen ones - seem like they should be included.

If it's primarily about JSON-serialization and structural equality, then there's not really any point in making them immutable - they could just be mutable JSON-serializable primitives.

It's worth looking into https://github.com/sebmarkbage/ecmascript-shallow-equal and why it was withdrawn - it seems like it might run into the same concerns.

mheiber: As I said, in my interpretation the proposal has multiple, related goals. And there may be a tradeoff between orthogonality and usefulness in this case.

Thanks for the link to the shallow equality proposal. That proposal seems to have been withdrawn for unrelated reasons: https://github.com/sebmarkbage/ecmascript-shallow-equal#status-of-this-proposal. In the shallow-equal proposal, the semantics weren't fully specced (for perf reasons), but vendors were afraid to implement because they sensed they would be locked in to certain implementations for backwards-compatibility.

mheiber:
related: https://github.com/rricard/proposal-const-value-types/issues/45

README.md
```diff
@@ -137,6 +138,7 @@ Array.from(tuple).map(x => new MyClass(x))
 assert((#{} with .a = 1, .b = 2) === #{ a: 1, b: 2 });
 assert((#[ {} ] with [0].a = 1) === #[ { a: 1 } ]);
 assert((x = 0, #[ {} ] with [x].a = 1) === #[ { a: 1 } ]);
+assert.equal(#{ a: 1 }, Record.from(JSON.stringify(#{ a: 1 })));
```
rickbutton: should this be:

```
assert(#{ a: 1 }, Record.from(JSON.parse(JSON.stringify(#{ a: 1 }))));
```

i.e. the string should be converted to an object first via `JSON.parse`

littledan:
I like the Record.from API, and I think the JSON serialization behavior is good documentation (it falls out of other things in this readme), but I don't understand the motivation for the JSON round-tripping property. It seems really odd to me to exclude symbols, undefined and BigInt. Further, I was hoping that this proposal would give us the core data model that would expand to "value classes" including private fields, which would definitely not JSON round-trip. And finally, sorting the keys breaks JSON round-tripping. Why not have as a goal, a subset of this will JSON round-trip?

mheiber:
Thanks for your feedback, @littledan . I guess I never wrote *why* I thought JSON roundtripping would be a good goal:

- We get good structured clone support for free. Related to this issue: https://github.com/rricard/proposal-const-value-types/issues/45
- One of the performance things we're excited about is that it is a step toward being able to cheaply get data from JSON. Few roundtripping hazards just seem nice if we can have it without giving up too much. I have in mind use cases reading/writing ASTs to/from disk when doing incremental compilation, or saving and rehydrating Redux stores.

1. Why is it odd to exclude symbols? We couldn't come up with a use case.

2. The `undefined` thing is odd, I agree, and I hadn't thought of `BigInt`. Does BigInt work with Structured Clone?

3. Re expanding the core data model more: does allowing JSON roundtripping for Record and Tuple preclude having 'value classes' with private fields later? (real question, not sure what the implications are)

4. Can you explain the sorting issue?

PR #43, opened 2019-08-17
clarify equality

mheiber:
- Clarify what we what we currently know about how the [four equality algorithms of JS](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness) interact with `Record` and `Tuple`.

- Note that we have not yet decided on how insertion order interacts with equality.

- Update wording to make it clear that there are no boxed Records or Tuples.

close https://github.com/rricard/proposal-const-value-types/issues/20

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ assert(#[1] === #[1]);
 
 // abstract equality comparison
 assert(#{ a:  1} == #{ a: 1 });
-assert(#[1] === #[1]);
+assert(#[1] == #[1]);
 
 // SameValue0
 assert(new Map().set(#{ a: 1 }, true).get(#{ a: 1 }));
@@ -253,10 +253,6 @@ This is related to [Ordering of properties](#ordering-of-properties).
 # `Record` and `Tuple` Globals
 
 We add to the global namespace two objects that you can use to manipulate records and tuples. Those objects have multiple properties that are in line with how records and tuples behave.
-
-The new globals are not constructors, and are distinct from the `Record` and `Tuple` types.
-
-Unlike `Boolean` and `String`, there is no such thing as a boxed `Record` or `Tuple`. For example, `Object.assign("hello")` silently creates a mutable `String` instance, but `Object.assign(#{})` produces a `TypeError`, just like `Object.assign(undefined)`. 
 
 ## Instantiation and converting from non-const types
 
@@ -494,6 +490,14 @@ If const classes are standardized, features like [Temporal Proposal](https://git
 ## What is different with this proposal than with [previous attempts](https://github.com/sebmarkbage/ecmascript-immutable-data-structures)?
 
 The main difference is that this proposal has a proper assignment operation using `with`. This difference makes it possible to handle proper type support, which was not possible with the former proposal.
+
+## Are there boxed versions of `Record` and `Tuple`?
+
+Not sure yet! Two options:
+- `Object(record)` creates a `Record` instance where `Object(#{}) !== Object(#{})`. This is akin to how `Object(true)` creates a `Boolean` instance such that `Object(true) !== Object(true)`.
+- `Object(record)` returns `{}`. This is the same behavior as for `Object(null)` and `Object(undefined)`.
+
+> Thanks @ljharb for raising this question.
 
 ## Would those matters be solved by a library and operator overloading?
 
--- a/README.md
+++ b/README.md
@@ -208,6 +208,48 @@ record with ["a"]["b"] = 1
 
 The same runtime verification will apply. It is a `TypeError` when a value inside a `Record` or `Tuple` is updated with a non-value type.
 
+# Equality
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -208,9 +208,55 @@ record with ["a"]["b"] = 1
...
+
+This is related to [Ordering of properties](#ordering-of-properties).
+
+# `Record` and `Tuple` Globals
+
+We add to the global namespace two objects that you can use to manipulate records and tuples. Those objects have multiple properties that are in line with how records and tuples behave.
+
+The new globals are not constructors, and are distinct from the `Record` and `Tuple` types.
+
+Unlike `Boolean` and `String`, there is no such thing as a boxed `Record` or `Tuple`. For example, `Object.assign("hello")` silently creates a mutable `String` instance, but `Object.assign(#{})` produces a `TypeError`, just like `Object.assign(undefined)`.
```
ljharb: fwiw, this is unacceptable to me - every non-nullish primitive in the language *must* have a boxed form. This is why `Object(Symbol())` and `Object(3n)` both work.

This is an axiom of the language, one that must not be broken.

mheiber: I wasn't aware of this invariant. Why is it important? Are you suggesting an alternative?

mheiber: Thanks for this feedback and for teaching me about `Object(thing)` which I hadn't heard of before.

## Two questions:

1. What is the invariant you're referring to and why is it important?
2. What alternative are you proposing? Are there use cases in favor of the alternative?

Re (1), it's unclear to me what invariant you're referring to.
- `Object(undefined)` and `Object(null)` both `ValueError` (https://www.ecma-international.org/ecma-262/6.0/#sec-toobject). I'd expect a `ValueError` for `Object(record)` as well.
- The `Symbol` example you bring up is interesting. Unlike the case with `boolean`'s evil twin `Boolean`, I don't think `Symbol` is not an evil twin. `Symbol()` is equivalent to `Object(Symbol())` as far as I can tell (from experimentation and spec reading). 

Re (2):

Are you proposing that
- Option A: `Object(#{ a: 1 })` creates a mutable Record. Note: I can't think of a use case for such a thing, since the whole point of records is that they are kind of like objects but immutable.
> I expect things like `Object(true)` were a mistake: why would anyone want a mutable boolean that doesn't even do the job of a boolean? `assert(Object(true) !== Object(true))`.

- Option B: That `Object(#{ a: 1})` creates an `Object` with the same keys and values as the record? That sounds good!  Would that become the recommended way of creating objects from records? Then we don't need the `Object.from` [that I proposed](https://github.com/rricard/proposal-const-value-types/pull/42).

ljharb: For one, there is code all over the web that checks if something is an object with `Object(x) === x` (or, a primitive with `Object(x) !== x`). That can't ever throw.

`Symbol()` is not equivalent to `Object(Symbol())`; the former is not instanceof Symbol, the latter is.

I'm not suggesting a mutable Record; I'm suggesting that it'd create a record that was `instanceof Record` (which the primitive form must not be).

mheiber: It sounds like you're suggesting that these two types of things will live side-by-side:
- A non-primitive, immutable Record
- A primitive, immutable Record

Will these things be equal? How can they be equal if one is a primitive and one is not?

-----

Re "symbol() is not equivalent to Object(Symbol())". It is!
I was surprised too, but read the spec and tried it out in Chrome.

```js
Object(Symbol()) instanceof Symbol // true
```


-------

You said that `Object(x) !== x` should never throw. That sounds like a good invariant to preserve!
What I wrote above was wrong, `Object(null)` and `Object(undefined)` do not throw, they return `{}`. So there's precedent: It's not like there is a `Null` type. Could we do something like that for `Record`? `Object(#{ a: 1 }) // returns {}`

ljharb:
The boxed form, like every object, must only ever be equal to itself.

README.md
ljharb: Also, that’s a special legacy case, because null and undefined can’t be boxed - i don’t think it’d be good to widen its usage.

ljharb: Additionally, no primitives are an instanceof anything - i don’t think that should change here either.

mheiber:
I'm starting to like `Object(#{ a: 1 }) // {}`.
When all options are bad, we must pick the least bad option.

ljharb:
It’d still have to be special, not a normal object - there’d have to be a way to toPrimitive it and extract the underlying Record (another axiom is that all types have at least one unforgeable brand-checking method)

README.md
mheiber: Thanks @ljharb!
Agree we'll probably need boxing.
Since this PR is more about equality than about boxing, I
- changed the PR so it doesn't express an opinion on whether we should box
- added the question about boxing to the FAQ so we can spec out any boxed objects when we get to a later proposal stage

mheiber:
Updated the PR to not express an opinion on boxing. Moved the boxing question to FAQ. Thanks @ljharb

Issue #41, opened 2019-08-16
Should `with` even be a part of this proposal?


TehShrike:
Continuing a thread from #1 – is there a compelling reason to include `with` in this proposal?

I don't believe this proposal needs the `with` syntax to be viable. I also doubt that the proposed `with` syntax needs to be specific to immutable data structures – `someObject with .b=5` would be handy as sugar for `Object.assign({}, someObject, { b: 5 })`.

Context:
- https://github.com/rricard/proposal-const-value-types/issues/1#issuecomment-502725556
- https://github.com/rricard/proposal-const-value-types/issues/1#issuecomment-521263307
- https://github.com/rricard/proposal-const-value-types/issues/37#issuecomment-522159428

ceigey:
I feel like having the two together is a good idea; otherwise, you might end up with a situation where, despite immutable data structures existing in the runtime, people are using good ol fashioned pretend-immutable data structures with libraries like Immer for the better developer ergonomics when making deep changes instead.

But, I do also feel like having something akin to Immer built in to JS is a great idea on its own too, and this repurposing of the `with` keyword provides that.

Speaking of deep property access, another question is what happens to `?.` and when the path the mutation is being made at doesn't exist; could `?.` be used to automatically fill in said path? And if so, do we assume that tuples are secretly a subset of records and records can be made from tuples?

So... if the proposal was split up, I wonder how deep their shared concerns would be; but evidently there's already other proposals that are completely separate that would have to be considered too.

TehShrike:
[zenflow](https://github.com/rricard/proposal-const-value-types/issues/1#issuecomment-499595272) brought up that the spread syntax already exists and would accomplish the same thing without adding a new language feature.

ceigey:
Spread syntax is ergonomic at one level deep, but after that, while perfectly functional, the syntax is not ideal and requires a fair bit of repetition.

Consider:

```javascript
let myRec = #{
  user: {
    address: {
      street: 'First St',
      //...
    }
    //...
  }
  //...
}

let newRecord = #{
  ...myRec,
  user: {
    ...myRec.user,
    address: {
      ...myRec.user.address,
      street: 'Second Ave'
    }
  }
}
```

The update would be better as something like:
```javascript
myRec.user.address.street = 'Second Ave';
```
... but that won't fly because we're dealing with immutable objects and the assignment operator can't "return" a value to assign to something else.

[...]

Issue #19, opened 2019-06-07
Have you considered using lenses for updates?


isiahmeadows:
<details>
<summary>For those who aren't familiar with lenses:</summary>

I know this sounds a bit like I'm getting ready to suggest some obscure functional programming concept, but I'm not.

Lenses are simple `{get(o): v, set(o, v): o}` pairs. They are really nice for operating on nested data without having to deal with all the boilerplate of the surrounding context. They are easy to write and easy to compose.

```js
// Gets and sets `b` inside of `a`
function nest(a, b) {
    return {
        get: o => b.get(a.get(o)),
        set: (o, v) => a.set(o, b.set(a.get(o), v)),
    }
}

// Updates a value inside a view
function modify(o, lens, func) {
    return lens.set(o, func(lens.get(o)))
}

// Updates a value inside a view path
// Called as `modifyIn(o, func, ...lenses)`
function modifyIn(o, func, lens, ...rest) {
    if (rest.length === 0) return modify(o, lens, func)
    return modifyIn(o, (v) => modify(v, lens, func), ...rest)
}
```

That `modify` operation is really where all the power in lenses lie, not really the `get` or `set`.

**Edit:** Forgot to double-check some variable names.
**Edit 2:** Fix yet another bug. Also, make it clearer which edits apply to this section.

-------
</details>

**Edit:** clarity, alter precedence of function call vs lens
**Edit 2:** Update to align with the current proposal

Lenses are pretty powerful and concise, and they provide easy, simple sugar over just direct, raw updates. But it kind of requires revising the model a bit:

- Getting properties could be something like `object.@lens`
- Setting properties could be something like `object with .@foo.@bar.@baz = value, ...`
- Updating properties could be something like `object with .@foo.@bar.@baz prev => next, ...`
- Sugar would exist for properties, using `.prop` instead of `.@lens`
- Sugar would exist for indices, using `[key]` instead of `.@lens`
- In each of these, you can do stuff like `.@foo(1, 2, 3).@bar("four", "five", "six")` and so on. The functions would be applied eagerly before actually evaluating updates. However, member expressions like `@(foo.bar)` need parenthesized.
- Yes, you can merge this all, like in `object with .@foo.bar[baz] prev => next, ...`.

Of course, I'm not beholden to the syntax here, and I'd like to see something a little more concise.

For a concrete example, consider this:

```js
const portfolio = #[
    { type: "bond", balance: 129.46 },
    { type: "bond", balance: 123.54 },
[...]

PR #49, opened 2019-09-09
Remove with from the proposal

rricard:
This is still contestable at that point but we are edging to removing the `with` part of the proposal to make it easier to reason with for the time being. As a followup, we should have an orthogonal proposal that would introduce that feature everywhere in the language (not only for const value types)

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ assert(record2 === #{ a: 1, c: 3, b: 5 });
 ```js
 const tuple1 = #[1, 2, 3];
 
-const tuple2 = #[1, ...tuple2];
+const tuple2 = #[1, ...tuple1];
 
 assert(tuple2 === #[1, 1, 2, 3]);
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -56,58 +56,30 @@ const record1 = #{
...
+assert(record2 === #{ a: 1, c: 3, b: 5 });
 ```
 
 #### Simple `Tuple`
 
 ```js
 const tuple1 = #[1, 2, 3];
 
-const tuple2 = tuple1 with [0] = 2;
+const tuple2 = #[1, ...tuple2];
```
chicoxyzzy:
```suggestion
const tuple2 = #[1, ...tuple1];
```

rricard: thanks good catch

rricard:
@robpalme I just added the .pop()/.push()/.with() examples that needed to be covered with the `with` keyword removal from the proposal

mheiber:
would close:
- https://github.com/rricard/proposal-const-value-types/issues/41
- https://github.com/rricard/proposal-const-value-types/issues/19

PR #50, opened 2019-09-10
Fix: true/false -> boolean

robpalme:
cc @ljharb


ljharb:
Why close?

robpalme:
Robin fixed it already. Thanks for identifying.

PR #51, opened 2019-09-10
Add bigint as permitted type

robpalme:


rricard:
We'll probably add Temporal as well later when it progresses on I imagine

littledan:
@rricard I doubt that; Temporal will be objects, not primitives.

rricard:
That's a bit of a shame since their API strongly hints they could be  treated as immutable values.

However that makes sense. Maybe if we end up adding const classes one day we could have const temporals...

littledan:
The API is very clear that they are objects. I'm a bit skeptical that it would be worth it to go back and make primitive versions of things that are already available as objects. And I strongly believe we shouldn't block other proposals on value classes, which are a ways out.

littledan:
CSSTOM already thought about doing this particular delay, and decided not to. I think we should stick to that pattern.

rricard:
I'm not saying we should backtrack and change them, I'm just saying at some point in the future we could see some const counterparts as an external library or in the language, but that's a long shot...

PR #55, opened 2019-09-10
order with the default comparator

rickbutton:
cc @ljharb @rricard

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ This part is an **open question**. We will gather additional feedback before dec
 ### Sorted Order (option 1)
 
 When the properties of a `Record` or `Tuple` are enumerated, their keys are enumerated in *sorted order*, where
-*sorted order* effectively means "according to the sort order defined by `Array.prototype.sort`".
+*sorted order* effectively means "according to the sort order defined by `Array.prototype.sort` with the default comparator".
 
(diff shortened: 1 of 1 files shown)


littledan:
I don't think the default comparator is exactly what we want here. Don't we want all of the indexed properties to come first, like in a normal object? Also, doesn't the default comparator throw on Symbols, which may be present as property keys?

rickbutton:
@littledan thats a good question! The proposal originally mimic-ed the existing object behavior of indexed properties coming first, but I simplified it. I have no opinion either way, I imagine your argument for it is consistency with how objects work?

Also very good point, the default comparator will throw for symbols. This isn't a problem for `Object.keys` or like `for...in` in general, because symbols don't appear in that enumerated list of properties. I imagine it would stay the same for records? If not, what behavior would you expect? Would symbols appear in the list of properties according to their description string?

ljharb:
Perhaps, instead of array sort, the order of Reflect.ownKeys should be used?

littledan:
@ljharb I think that would be insertion order (modulo the integer key sorting thing I mentioned earlier), which is exactly what we don't have.

Let's discuss this in an issue. I think this could be a bit of a more detailed discussion. We really just need something well-defined.

PR #58, opened 2019-09-17
"record keys are not ordered

mheiber:
Update our explainer to reflect that record keys are not ordered

We determined that having a notion of equality for records
that is dependent on insertion order would have poor
ergonomics and would not buy us any performance gain.

This PR does *not* resolve the question of what the sort function will be for the return value of `Record.keys`. @rickbutton is working on that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ assert(Object(#{ a: 1 }) !== Object(#{ a: 1 }));
 assert(Object(#[1, 2]) !== Object(#[1, 2]));
 ```
 
-Insertion order does not affect equality of records:
+Insertion order of record keys does not affect equality of records:
 
 ```js
 assert(#{ a: 1, b: 2 } === #{ b: 2, a: 1 });
--- a/README.md
+++ b/README.md
@@ -149,11 +149,22 @@ Instances of `Record` and `Tuple` are immutable, so their equality works like th
 
 ```js
 assert(#{ a: 1 } === #{ a: 1 });
-assert(Object(#{ a: 1 }) !== Object(#{ a: 1 }));
-assert({ a: 1 } !== { a: 1 });
+assert(#[1, 2] === #[1, 2]);
 ```
 
-This is distinct from how equality works for JS objects: `assert({} !== {});`. Strict comparison of objects will observe that each object is distinct.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -149,11 +149,22 @@ Instances of `Record` and `Tuple` are immutable, so their equality works like th
...
+This is distinct from how equality works for JS objects. Strict comparison of objects will observe that each object is distinct: 
+
+```js
 assert({ a: 1 } !== { a: 1 });
+assert(Object(#{ a: 1 }) !== Object(#{ a: 1 }));
+assert(Object(#[1, 2]) !== Object(#[1, 2]));
 ```
 
-This is distinct from how equality works for JS objects: `assert({} !== {});`. Strict comparison of objects will observe that each object is distinct.
+Insertion order does not affect equality of records:
```
ljharb: this implies that it does affect equality of tuples; can an example be provided?

mheiber: rewrote to "Insertion order of record keys does not affect equality of records"

littledan:
LGTM. Let's merge this PR.

README.md
```diff
@@ -149,11 +149,22 @@ Instances of `Record` and `Tuple` are immutable, so their equality works like th
...
+This is distinct from how equality works for JS objects. Strict comparison of objects will observe that each object is distinct: 
+
+```js
 assert({ a: 1 } !== { a: 1 });
+assert(Object(#{ a: 1 }) !== Object(#{ a: 1 }));
+assert(Object(#[1, 2]) !== Object(#[1, 2]));
 ```
 
-This is distinct from how equality works for JS objects: `assert({} !== {});`. Strict comparison of objects will observe that each object is distinct.
+Insertion order of record keys does not affect equality of records:
```
ljharb: I’d still like an example of ordering mattering for tuples here.

rickbutton: @mheiber ^

rickbutton: or alternatively I can add the example, up to you @mheiber

mheiber: Sure @rickbutton

I was not expecting insertion order to matter for tuples.

That is, I would expect `#[...#[1]]` to equal `#[1]`

ljharb: That’s not insertion order tho.

It’d be, would [1, 2] equal [2, 1]?

PR #67, opened 2019-09-20
Add performance expectations

rricard:


README.md
```diff
@@ -368,6 +383,19 @@ weakMap.set(record, true);
 
 # FAQ
 
+## What are the performance expectations of those Data Structures?
+
+This proposal in itself does not put any performance guarantees and does not require specific optimizations on the implementers. It is however built in a way that some performance optimizations can be done in most cases if implementers choose to do so.
```
ljharb: Map and Set impose specific requirements on implementations - namely, that retrieval must be sublinear as the size of the collection increases - why not do the same here?

littledan: This is more like strings. it's not really possible to guarantee both constant time concatenation and indexing at once, but implementations try to make it kind of work anyway. We should explain this explicitly. This could also be a good place to paste the text I wrote in an issue about this topic.

rricard:
@littledan I just copied your writeup from https://github.com/rricard/proposal-const-value-types/issues/2#issuecomment-533084421 in the PR

README.md
rricard: just pasted the text you wrote, I just had to retrieve it

PR #68, opened 2019-09-20
added small explainer for deep spread follow up proposal

rickbutton:
small write up of a possible follow up proposal solving the "deep update" problem via deep spread syntax.

### DIFF
--- a/Deep-Spread.md
+++ b/Deep-Spread.md
@@ -1,0 +1,92 @@
+All credit for this idea goes to Lee Byron @leebyron!
+
+The additional syntax below can be introduced as a follow up proposal.
+
+# Deep Spread for Records
+
+Currently, only spread syntax can be used to manipulate records, via creating a new record
+with the same properties as the old record, overwriting with the properties specified in the new record literal.
+
+If deep path updates are required (updating a value nested deeply inside a record), then the syntax is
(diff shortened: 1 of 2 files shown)


Deep-Spread.md
```diff
@@ -0,0 +1,92 @@
...
+Question! What happens if the deep path does not exist in the value that is spread?
+
+Example:
+
+```js
+const one = { a: 1 };
+
+const two = { b.c: 2, ...one };
+
+// Does this fail, or create a record that looks like:
```
littledan: I would say fail, otherwise it becomes difficult to know whether a Record or a Tuple is created when the property key is numeric.

Deep-Spread.md
```diff
@@ -0,0 +1,92 @@
+All credit for this idea goes to Lee Byron @leebyron!
+
+The additional syntax below can be introduced as a follow up proposal.
+
+# Deep Spread for Records
```
littledan: I wouldn't call this "deep spreads", more like, "deep property definitions". It's just used *in conjunction with* spread.

Deep-Spread.md
rickbutton: ah yes, good point!

PR #79, opened 2019-10-10
typeof tuple is "tuple"

littledan:


rickbutton:
good catch @littledan

PR #80, opened 2019-10-15
Stage 1 Comments: All nested usages of records & tuples needs to use #

rricard:


ljharb:
Alternatively, objects and arrays could be auto coerced into records and tuples.

rricard:
Well, I'd like to consult the committee before going into that rabbit hole, for now let's play the "safe" way and require # everywhere

rickbutton:
+1, I'm not opposed to having the discussion about auto-coercion, but at the very least the syntax change should be made.

PR #81, opened 2019-10-15
New past tense methods

rricard:
Instead of push/pop/etc... be overridden, we keep them the same but create new methods in the past sense

### DIFF
--- a/NS-Proto-Appendix.md
+++ b/NS-Proto-Appendix.md
@@ -78,9 +78,17 @@ Tuple.of(1,2,3); // #[1, 2, 3]
 
 ## `Tuple.prototype.pop()`
 
+Returns the last value of the tuple.
+
+## `Tuple.prototype.popped()`
+
 Returns a `Tuple` identical to the original `Tuple` except the last element is removed.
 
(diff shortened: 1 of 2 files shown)


ljharb:
So would pop and push mutate? If not, they shouldn’t be present.

rricard:
In the appendix the push is a no-op, and pop will return the last elem.

I do still want to keep them for now just in case we're able to possibly add pushed and popped and the others to Array.prototype.

ljharb:
Why would push ever be a noop? Why would pop ever not pop the item - ie, remove it?

rricard:
effectively having push as a no-op is not super useful. having pop as giving you the last element might actually end up being useful unless we add a .last()

ljharb:
The name "pop" is destructive and mutates. It seems reasonable to advance `.last()` separately if that would be useful, but an immutable value should not have apparent mutator methods on it.

rricard:
cc. @rickbutton let's have a chat about this this week

papb:
If a method is a noop, I think it simply shouldn't exist instead. This would prevent bugs.

PR #89, opened 2019-12-24
Improvements covering a few ongoing issues

### LINKED ISSUE #40: examples of non-assignment/update member expression operations?
All the examples in the readme are with updates, I assume:

```mjs
let a = #{x: 1};
assert(a.x === 1);
let b = #[2];
assert(b[0] === 2);
```

Are the expected syntax for reading fields in this manner and is intended to work.

rricard:
This change fixes #40, fixes #74, fixes #88.

If anything is too controversial, I can isolate a commit and discuss it further in an another PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,8 @@ assert(record1.a === 1);
 assert(record1["a"] === 1);
 assert(record1 !== record2);
 assert(record2 === #{ a: 1, c: 3, b: 5 });
+assert(record1?.a === 1);
+assert(record1?.d ?? 5 === 5);
 ```
 
 #### `Tuple`
--- a/README.md
+++ b/README.md
@@ -155,13 +155,7 @@ At runtime, attempting to create a `Record` with a key that is not a `string` or
 
 At runtime, it is a `TypeError` to add a value to a `Record` or `Tuple` of any type except the following: `Record`, `Tuple`, `string`, `number`, `symbol`, `boolean`, `bigint`, `undefined`, or `null`.
 
-However whenever the engine can figure out at parse time that something is wrong, it should fail at that moment:
-
-```js
-const r1 = #{ obj: {} }; // fails at parse time because an object declaration in a record is obviously going to be an error
-const r2 = #{ obj: true ? {} : #{} }; // will fail at runtime as it's not obvious from the parser's perspective that we should fail
-const r3 = #{ method() {} }; // fails at parse time
-```
+If you try to use a method definitions as part of a Record, the same behavior is expected, it should fail at runtime (even if we could fail at compile time).
 
 # Equality
 
@@ -343,10 +337,6 @@ JSON.stringify(#[true, #{ a: #[1, 2, 3] }]); // '[true,{"a":[1,2,3]}]'
 We propose to add `JSON.parseImmutable` so we can extract a Record/Tuple type out of a JSON string instead of an Object/Array.
 
 The signature of `JSON.parseImmutable` is identical to `JSON.parse` with the only change being in the return type that is now a Record or a Tuple.
-
-## JSON.stringify
-
-Stringifying a record or a tuple should output object or array syntax that corresponds recursively.
 
 ## `Record` prototype
 
--- a/README.md
+++ b/README.md
@@ -156,9 +156,9 @@ At runtime, it is a `TypeError` to add a value to a `Record` or `Tuple` of any t
 However whenever the engine can figure out at parse time that something is wrong, it should fail at that moment:
 
 ```js
-const r1 = { obj: {} }; // fails at parse time because an object declaration in a record is obviously going to be an error
-const r2 = { obj: true ? {} : #{} }; // will fail at runtime as it's not obvious from the parser's perspective that we should fail
-const r3 = { method() {} }; // fails at parse time
+const r1 = #{ obj: {} }; // fails at parse time because an object declaration in a record is obviously going to be an error
+const r2 = #{ obj: true ? {} : #{} }; // will fail at runtime as it's not obvious from the parser's perspective that we should fail
+const r3 = #{ method() {} }; // fails at parse time
 ```
 
 # Equality
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ assert(record1["a"] === 1);
 assert(record1 !== record2);
 assert(record2 === #{ a: 1, c: 3, b: 5 });
 assert(record1?.a === 1);
+assert(record1?.d === undefined);
 assert(record1?.d ?? 5 === 5);
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -151,6 +153,14 @@ At runtime, attempting to create a `Record` with a key that is not a `string` or
 
 At runtime, it is a `TypeError` to add a value to a `Record` or `Tuple` of any type except the following: `Record`, `Tuple`, `string`, `number`, `symbol`, `boolean`, `bigint`, `undefined`, or `null`.
 
+However whenever the engine can figure out at parse time that something is wrong, it should fail at that moment:
+
+```js
+const r1 = { obj: {} }; // fails at parse time because an object declaration in a record is obviously going to be an error
+const r2 = { obj: true ? {} : #{} }; // will fail at runtime as it's not obvious from the parser's perspective that we should fail
+const r3 = { method() {} }; // fails at parse time
```
dcporter: (I think r1 and r3 are missing #s? Will delete if I missed something)

README.md
```diff
@@ -71,6 +71,8 @@ assert(record1.a === 1);
 assert(record1["a"] === 1);
 assert(record1 !== record2);
 assert(record2 === #{ a: 1, c: 3, b: 5 });
+assert(record1?.a === 1);
+assert(record1?.d ?? 5 === 5);
 ```
```
ljharb: might be too to show an example where the optional chain is called on a nullish LHS

README.md
```diff
@@ -151,6 +153,14 @@ At runtime, attempting to create a `Record` with a key that is not a `string` or
 
 At runtime, it is a `TypeError` to add a value to a `Record` or `Tuple` of any type except the following: `Record`, `Tuple`, `string`, `number`, `symbol`, `boolean`, `bigint`, `undefined`, or `null`.
 
+However whenever the engine can figure out at parse time that something is wrong, it should fail at that moment:
+
+```js
+const r1 = { obj: {} }; // fails at parse time because an object declaration in a record is obviously going to be an error
```
ljharb:
```suggestion
const r1 = #{ obj: {} }; // fails at parse time because an object declaration in a record is obviously going to be an error
```

README.md
```diff
@@ -151,6 +153,14 @@ At runtime, attempting to create a `Record` with a key that is not a `string` or
...
+const r2 = { obj: true ? {} : #{} }; // will fail at runtime as it's not obvious from the parser's perspective that we should fail
```
ljharb:
```suggestion
const r2 = #{ obj: true ? {} : #{} }; // will fail at runtime as it's not obvious from the parser's perspective that we should fail
```

README.md
ljharb:
```suggestion
const r3 = #{ method() {} }; // fails at parse time
```

rricard: Good catch, all of them are missing

README.md
rricard: yes, makes sense thanks!

ljharb: Note that this still isn’t here; i mean something that’s the equivalent of `null?.something` or `undefined?.something` - without that the “optional” part is useless.

rricard: oh sorry I did not get it like this, so something like `record1.d?.a === undefined`? Otherwise I could use your help to write that example because unfortunately I feel like I'm missing something obvious.

ljharb: Yes, exactly - basically, something nullish in the record, that someone tries to optionally chain off of.

README.md
```diff
@@ -151,6 +156,14 @@ At runtime, attempting to create a `Record` with a key that is not a `string` or
 
 At runtime, it is a `TypeError` to add a value to a `Record` or `Tuple` of any type except the following: `Record`, `Tuple`, `string`, `number`, `symbol`, `boolean`, `bigint`, `undefined`, or `null`.
 
+However whenever the engine can figure out at parse time that something is wrong, it should fail at that moment:
```
rickbutton: are we sure about this? This runs counter to my [babel implementation](https://github.com/babel/babel/pull/10865), where any expressions inside a record are valid at parse time (including method definitions), but fail at runtime.

README.md
```diff
@@ -326,6 +339,12 @@ JSON.stringify(#{ a: #[1, 2, 3] }); // '{"a":[1,2,3]}'
 JSON.stringify(#[true, #{ a: #[1, 2, 3] }]); // '[true,{"a":[1,2,3]}]'
 ```
 
+## JSON.parseImmutable
```
rickbutton: should we also clarify what `JSON.stringify` does? Does it output JSON object syntax, or fail?

rickbutton: presumably, output JSON object syntax is the correct answer, since it can then be consumed again by `JSON.parseImmutable`, but want to verify your thoughts.

README.md
rricard: Discussions in #88 seems to indicate that we would prefer parse time errors if possible. The Babel implementation does not need to be fully compliant with that or could eventually be improved later.

README.md
rricard: Yes it should output JSON object syntax

README.md
littledan: I'm confused. I think the conclusion on #88 was runtime errors.

rricard: I reread it and yes it was indeed the contrary

README.md
```diff
@@ -155,13 +155,7 @@ At runtime, attempting to create a `Record` with a key that is not a `string` or
...
 At runtime, it is a `TypeError` to add a value to a `Record` or `Tuple` of any type except the following: `Record`, `Tuple`, `string`, `number`, `symbol`, `boolean`, `bigint`, `undefined`, or `null`.
 
-However whenever the engine can figure out at parse time that something is wrong, it should fail at that moment:
-
-```js
-const r1 = #{ obj: {} }; // fails at parse time because an object declaration in a record is obviously going to be an error
-const r2 = #{ obj: true ? {} : #{} }; // will fail at runtime as it's not obvious from the parser's perspective that we should fail
-const r3 = #{ method() {} }; // fails at parse time
-```
+If you try to use a method definitions as part of a Record, the same behavior is expected, it should fail at runtime (even if we could fail at compile time).
```
littledan: Not sure why you removed the examples

README.md
```diff
@@ -344,10 +338,6 @@ We propose to add `JSON.parseImmutable` so we can extract a Record/Tuple type ou
 
 The signature of `JSON.parseImmutable` is identical to `JSON.parse` with the only change being in the return type that is now a Record or a Tuple.
 
-## JSON.stringify
```
littledan: Not sure why this was removed either; it's good to know, even if there's zero spec text corresponding to it.

rickbutton: It was already specified somewhere else in the file, so this was a duplicate.

PR #124, opened 2020-05-28
Add back Record.isRecord

littledan:
This was included before, and turns out to be needed in the context of #71

### DIFF
--- a/NS-Proto-Appendix.md
+++ b/NS-Proto-Appendix.md
@@ -4,6 +4,10 @@
 
 Converts shallowly an object to a record. If the object has a non-const value, a TypeError will be thrown.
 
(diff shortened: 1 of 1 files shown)


littledan:
cc @ljharb

Jack-Works:
Why not just typeof

rricard:
typeof will detect the primitive type (record, tuple) but not the primitive type boxed into their boxing objects (Record, Tuple). It is an edge case but it could be useful to detect it.

ljharb:
More specifically, it’s an axiom for every object value (except Error, for legacy reasons), including boxed primitives, that there must be a cross-realm mechanism to distinguish it.

PR #130, opened 2020-06-04
Switch to SameValueZero comparison semantics

littledan:
One possible solution for #65

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,21 +266,24 @@ Object.keys(#{ a: 1, b: 2 })  // ["a", "b"]
 Object.keys(#{ b: 2, a: 1 })  // ["a", "b"]
 ```
 
-If their structure and contents are deeply identical, then `Record` and `Tuple` values considered equal according to all of the equality operations: `Object.is`, `==`, `===`, and the internal SameValueZero algorithm used for Maps and Sets. See further discussion in [#65](https://github.com/rricard/proposal-const-value-types/issues/65).
+If their structure and contents are deeply identical, then `Record` and `Tuple` values considered equal according to all of the equality operations: `Object.is`, `==`, `===`, and the internal SameValueZero algorithm used for Maps and Sets.
+
+The `==`, `===` and SameValueZero algorithms consider the Numbers `0` and `-0` to be equal to each other. This equality filters up through Records and Tuples, whereas they are not considered the same by `Object.is`. Overall, `==` and `===` in Records and Tuples recursively performs the SameValueZero algorithm recursively on contents, in order to unify the two values, while maintaining other desirable properties of comparison operators. See further discussion in [#65](https://github.com/rricard/proposal-const-value-types/issues/65).
 
 ```js
 assert(#{ a:  1 } === #{ a: 1 });
 assert(#[1] === #[1]);
 
-assert(#{ a: -0 } !== #{ a: +0 });
-assert(#[-0] !== #[+0]);
+assert(#{ a: -0 } === #{ a: +0 });
+assert(#[-0] === #[+0]);
 assert(#{ a: NaN } === #{ a: NaN });
 assert(#[NaN] === #[NaN]);
 
-assert(#{ a: -0 } != #{ a: +0 });
-assert(#[-0] != #[+0]);
+assert(#{ a: -0 } == #{ a: +0 });
+assert(#[-0] == #[+0]);
 assert(#{ a: NaN } == #{ a: NaN });
 assert(#[NaN] == #[NaN]);
+assert(#[1] != #["1"]);
 
 assert(!Object.is(#{ a: -0 }, #{ a: +0 }));
 assert(!Object.is(#[-0], #[+0]));
--- a/README.md
+++ b/README.md
@@ -266,9 +266,13 @@ Object.keys(#{ a: 1, b: 2 })  // ["a", "b"]
 Object.keys(#{ b: 2, a: 1 })  // ["a", "b"]
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -290,6 +293,7 @@ assert(Object.is(#[NaN], #[NaN]));
 // SameValueZero
```
rricard, line 293: I think we can get rid of this comment:
```suggestion
```

README.md
```diff
@@ -266,21 +266,24 @@ Object.keys(#{ a: 1, b: 2 })  // ["a", "b"]
 Object.keys(#{ b: 2, a: 1 })  // ["a", "b"]
 ```
 
-If their structure and contents are deeply identical, then `Record` and `Tuple` values considered equal according to all of the equality operations: `Object.is`, `==`, `===`, and the internal SameValueZero algorithm used for Maps and Sets. See further discussion in [#65](https://github.com/rricard/proposal-const-value-types/issues/65).
+If their structure and contents are deeply identical, then `Record` and `Tuple` values considered equal according to all of the equality operations: `Object.is`, `==`, `===`, and the internal SameValueZero algorithm used for Maps and Sets.
+
+The `==`, `===` and SameValueZero algorithms consider the Numbers `0` and `-0` to be equal to each other. This equality filters up through Records and Tuples, whereas they are not considered the same by `Object.is`. Overall, `==` and `===` in Records and Tuples recursively performs the SameValueZero algorithm recursively on contents, in order to unify the two values, while maintaining other desirable properties of comparison operators. See further discussion in [#65](https://github.com/rricard/proposal-const-value-types/issues/65).
```
rricard, line 272: I feel like we can just add a clarifying list here:

```suggestion

- `===` and `==` both perform a recursive SameValueZero on each side
- `Object.is` performs a recursive `Object.is` on each side
- SameValueZero is still SameValueZero and is performed recursively

```

README.md
littledan: I guess the comment exists to indicate that the SameValueZero algorithm is used by Maps (since that's not well-known to all JS programmers).

rricard: Ok so maybe we could factor [this list](https://github.com/tc39/proposal-record-tuple/pull/130#discussion_r435213074) inside of the examples as comments, that should dissipate all confusions and I would visually like that better in the proposal

README.md
```diff
@@ -266,21 +266,24 @@ Object.keys(#{ a: 1, b: 2 })  // ["a", "b"]
...
 
 ```js
 assert(#{ a:  1 } === #{ a: 1 });
 assert(#[1] === #[1]);
```
rricard, line 276:
```suggestion

// `===` and `==` both perform a recursive SameValueZero on each side
```

README.md
rricard: disregard that one

README.md
```diff
@@ -266,21 +266,24 @@ Object.keys(#{ a: 1, b: 2 })  // ["a", "b"]
...
 assert(#{ a: NaN } === #{ a: NaN });
 assert(#[NaN] === #[NaN]);
 
-assert(#{ a: -0 } != #{ a: +0 });
-assert(#[-0] != #[+0]);
+assert(#{ a: -0 } == #{ a: +0 });
+assert(#[-0] == #[+0]);
 assert(#{ a: NaN } == #{ a: NaN });
 assert(#[NaN] == #[NaN]);
+assert(#[1] != #["1"]);
```
rricard, line 287:
```suggestion

//  `Object.is` performs a recursive `Object.is` on each side
```

README.md
rricard:
```suggestion
// SameValueZero is still SameValueZero and is performed recursively
```

README.md
```diff
@@ -266,30 +266,38 @@ Object.keys(#{ a: 1, b: 2 })  // ["a", "b"]
 Object.keys(#{ b: 2, a: 1 })  // ["a", "b"]
 ```
 
-If their structure and contents are deeply identical, then `Record` and `Tuple` values considered equal according to all of the equality operations: `Object.is`, `==`, `===`, and the internal SameValueZero algorithm used for Maps and Sets. See further discussion in [#65](https://github.com/rricard/proposal-const-value-types/issues/65).
+If their structure and contents are deeply identical, then `Record` and `Tuple` values considered equal according to all of the equality operations: `Object.is`, `==`, `===`, and the internal SameValueZero algorithm (used for comparing keys of Maps and Sets). They differ in terms of how `-0` is treated:
```
ljharb, line 269:
```suggestion
If their structure and contents are deeply identical, then `Record` and `Tuple` values considered equal according to all of the equality operations: `Object.is`, `==`, `===`, and the internal SameValueZero algorithm (used for comparing keys of Maps and Sets). They differ in terms of how `-0` is treated, and how NaN is treated:
```

README.md
```diff
@@ -266,30 +266,38 @@ Object.keys(#{ a: 1, b: 2 })  // ["a", "b"]
...
+- `Object.is` treats `-0` and `0` as unequal
+- `==`, `===` and SameValueZero treat `-0` with `0` as equal
```
ljharb, line 271:
```suggestion
- `Object.is` (`SameValue`) treats `-0` and `0` as unequal
- `==`, `===`, and `SameValueZero` treat `-0` with `0` as equal

- `Object.is` (`SameValue`) treat `NaN` and `NaN` as equal
- `==`, `===`, and `SameValueZero` treat NaN as unequal
```

littledan: So, it sounds like this section is unclear. It's trying to talk about how the operators work when dealing with data nested within Records and Tuples: The purpose of this section is to explain how we're making `#[NaN] === #[NaN]`.

littledan:
I'm going to land this as is, and we can consider clarifications in a follow-on PR. Let's leave discussion open in #65 but consider this the champion group's default option for Stage 2 going forward.

rickbutton:
great work on this @rricard @littledan

PR #131, opened 2020-06-09
Initial Record/Tuple standard library spec

rickbutton:
NOTE: This is an early and incomplete draft. Comments welcome.

### DIFF
--- a/spec/immutable-data-structures.html
+++ b/spec/immutable-data-structures.html
@@ -92,7 +92,6 @@
         <emu-alg>
           1. If NewTarget is not *undefined*, throw a *TypeError* exception.
           1. Let _elements_ be a List containing the arguments passed to this function.
-          1. Let _length_ to be the number of elements in _elements_.
           1. For each element _e_ of _elements_,
             1. If Type(_e_) is Object, throw a *TypeError* exception.
           1. Let _tuple_ be a new Tuple value whose [[Sequence]] is _elements_.
@@ -157,7 +156,6 @@
         <p>The *of* method takes any number of arguments, and performs the following steps:</p>
         <emu-alg>
           1. Let _elements_ be a List containing the arguments passed to this function.
-          1. Let _length_ to be the number of elements in _elements_.
           1. For each element _e_ of _elements_,
             1. If Type(_e_) is Object, throw a *TypeError* exception.
           1. Let _tuple_ be a new Tuple value whose [[Sequence]] is _elements_.
@@ -195,7 +193,7 @@
         <h1>Tuple.prototype.valueOf ( )</h1>
         <p>When the `valueOf` function is called, the following steps are taken:</p>
         <emu-alg>
-          1. Return ? thisTupleValue(*this* value)
+          1. Return ? thisTupleValue(*this* value).
         </emu-alg>
       </emu-clause>
       <emu-clause id="sec-tuple.prototype-@@toStringTag">
--- a/spec/immutable-data-structures.html
+++ b/spec/immutable-data-structures.html
@@ -47,25 +47,16 @@
           1. Perform ? RequireObjectCoercible(_iterable_).
           1. Let _fields_ be a new empty List.
           1. Let _iteratorRecord_ be ? GetIterator(_iterable_).
-          1. Let _stepsDefine_ be the algorithm steps defined in <emu-xref href="#sec-append-record-field-to-list-functions" title></emu-xref>.
-          1. Let _adder_ be ! CreateBuiltinFunction(_stepsDefine_, &laquo; &raquo;).
-          1. Perform ! AddEntriesFromIterable(_fields_, _iterable_, _adder_).
+          1. Let _adder_ be a new Abstract Closure with parameters (_key_, _value_) that captures _fields_ and performs the following steps when called:
+            1. If Type(_value_) is Object, throw a *TypeError* exception.
+            1. Let _field_ be { [[Key]]: _key_, [[Value]]: _value_ }.
+            1. Append _field_ to the end of list _fields_.
+          1. Perform ! AddEntriesFromIterable(`undefined`, _iterable_, _adder_).
           1. Return a new Record value whose [[Fields]] is _fields_.
         </emu-alg>
         <emu-note>
           <p>The parameter _iterable_ is expected to be an object that implements an @@iterator method that returns an iterator object that produces a two element array-like object whose first element is a value that will be used as a Map key and whose second element is the value to associate with that key.</p>
         </emu-note>
-      </emu-clause>
-      <emu-clause id="sec-append-record-field-to-list-functions">
-        <h1>AppendRecordFieldToList Functions</h1>
-        <p>An AppendRecordFieldToList function is an anonymous built-in function. When an AppendRecordFieldtoList function is called with arguments _key_ and _value_, the following steps are taken:</p>
-        <emu-alg>
-            1. Let _list_ be the *this* value.
-            1. Assert: Type(_list_) is List.
-            1. If Type(_value_) is Object, throw a *TypeError* exception.
-            1. Let _field_ be { [[Key]]: _key_, [[Value]]: _value_ }.
-            1. Append _field_ to the end of list _list_.
-        </emu-alg>
       </emu-clause>
       <emu-clause id="sec-record.isrecord">
         <h1>Record.isRecord ( _arg_ )</h1>
--- a/spec/immutable-data-structures.html
+++ b/spec/immutable-data-structures.html
@@ -323,7 +323,13 @@
         <p>The initial value of the *Tuple.prototype.lastIndexOf* is the *"lastIndexOf"* property of the intrinsic object %Array.prototype%.</p>
       </emu-clause>
       <emu-clause id="sec-tuple.prototype.entries">
-        <h1>Tuple.prototype.entries ()</h1>
+        <h1>Tuple.prototype.entries ( )</h1>
+        <p>When the *entries* method is called, it returns an iterator over the entries of the tuple.</p>
+        <p>The following steps are taken:</p>
+        <emu-alg>
+          1. Let _T_ be ? thisTupleValue(*this* value).
+          1. Return CreateTupleIterator(_T_, ~key+value~).
+        </emu-alg>
       </emu-clause>
       <emu-clause id="sec-tuple.prototype.every">
         <h1>Tuple.prototype.every</h1>
@@ -345,7 +351,13 @@
         <p>The initial value of the *Tuple.prototype.forEach* is the *"forEach"* property of the intrinsic object %Array.prototype%.</p>
       </emu-clause>
       <emu-clause id="sec-tuple.prototype.keys">
-        <h1>Tuple.prototype.keys ()</h1>
+        <h1>Tuple.prototype.keys ( )</h1>
+        <p>When the *keys* method is called, it returns an iterator over the keys of the tuple.</p>
+        <p>The following steps are taken:</p>
+        <emu-alg>
+          1. Let _T_ be ? thisTupleValue(*this* value).
+          1. Return CreateTupleIterator(_T_, ~key~).
+        </emu-alg>
       </emu-clause>
       <emu-clause id="sec-tuple.prototype.map">
         <h1>Tuple.prototype.map ()</h1>
@@ -382,12 +394,27 @@
       </emu-clause>
       <emu-clause id="sec-tuple.prototype.values">
         <h1>Tuple.prototype.values ()</h1>
+        <p>When the *values* method is called, it returns an iterator over the values of the tuple.</p>
+        <p>The following steps are taken:</p>
+        <emu-alg>
+          1. Let _T_ be ? thisTupleValue(*this* value).
+          1. Return CreateTupleIterator(_T_, ~value~).
+        </emu-alg>
       </emu-clause>
       <emu-clause id="sec-tuple.prototype-@@iterator">
         <h1>Tuple.prototype [ @@iterator ] ()</h1>
       </emu-clause>
       <emu-clause id="sec-tuple.prototype.with">
-        <h1>Tuple.prototype.with ()</h1>
+        <h1>Tuple.prototype.with ( _index_, _value_ )</h1>
+        <p>When the *with* method is called with two arguments, it returns a new tuple with the element at index _index_ replaced with value _value_.</p>
+        <emu-alg>
+          1. Let _T_ be ? thisTupleValue(*this* value).
+          2. Let _list_ be CreateListFromTuple(_T_).
+          3. Let _length_ be the length of list _list_.
+          4. If _index_ &lt; 0 or _index_ >= _length_, throw a *TypeError* exception.
+          5. Set _list_[_index_] to _value_.
+          6. Return ? CreateTupleFromList(_list_).
+        </emu-alg>
       </emu-clause>
     </emu-clause>
   </emu-clause>
--- a/spec/immutable-data-structures.html
+++ b/spec/immutable-data-structures.html
@@ -21,10 +21,11 @@
         <emu-alg>
           1. If NewTarget is not *undefined*, throw a *TypeError* exception.
           1. Let _obj_ be ? ToObject(_arg_).
-          1. Let _nameList_ be ? EnumerableOwnPropertyNames(_obj_, ~key~).
+          1. Let _props_ be ? EnumerableOwnPropertyNames(_obj_, ~key+value~).
           1. Let _fields_ be a new empty List.
-          1. For each element _name_ of _nameList_,
-            1. Let _value_ be ? GetV(_obj_, _name_).
+          1. For each element _prop_ of _props_,
+            1. Let _name_ be ? GetV(_obj_, `"0"`).
+            1. Let _value_ be ? GetV(_obj_, `"1"`).
             1. If Type(_value_) is Object, throw a *TypeError* exception.
             1. Let _field_ be { [[Key]]: _name_, [[Value]]: _value_ }.
             1. Append _field_ to the end of list _fields_.
--- a/spec/immutable-data-structures.html
+++ b/spec/immutable-data-structures.html
@@ -24,8 +24,8 @@
           1. Let _props_ be ? EnumerableOwnPropertyNames(_obj_, ~key+value~).
           1. Let _fields_ be a new empty List.
           1. For each element _prop_ of _props_,
-            1. Let _name_ be ! GetV(_obj_, `"0"`).
-            1. Let _value_ be ! GetV(_obj_, `"1"`).
+            1. Let _name_ be ! GetV(_prop_, `"0"`).
+            1. Let _value_ be ! GetV(_prop_, `"1"`).
             1. If Type(_value_) is Object, throw a *TypeError* exception.
             1. Let _field_ be the Record { [[Key]]: _name_, [[Value]]: _value_ }.
             1. Append _field_ to the end of list _fields_.
--- a/spec/immutable-data-structures.html
+++ b/spec/immutable-data-structures.html
@@ -517,7 +517,7 @@
           1. Let _T_ be ? thisTupleValue(*this* value).
           2. Let _list_ be CreateListFromTuple(_T_).
           3. Let _length_ be the length of list _list_.
-          4. If _index_ &lt; 0 or _index_ >= _length_, throw a *TypeError* exception.
+          4. If _index_ &lt; 0 or _index_ >= _length_, throw a *RangeError* exception.
           5. Set _list_[_index_] to _value_.
           6. Return ? CreateTupleFromList(_list_).
         </emu-alg>
--- a/spec/immutable-data-structures.html
+++ b/spec/immutable-data-structures.html
@@ -71,7 +71,9 @@
         <h1>Record.isRecord ( _arg_ )</h1>
         <p>The *isRecord* function takes one argument _arg_, and performs the following steps:</p>
         <emu-alg>
-          1. Return ? IsRecord(_arg_).
+          1. If Type(_arg_) is Record, return `true`.
+          1. If Type(_arg_) is Object and _arg_ has a [[RecordData]] internal slot, return `true`.
+          1. Return `false`.
         </emu-alg>
       </emu-clause>
       <emu-clause id="sec-record.prototype">
@@ -119,7 +121,9 @@
         <h1>Tuple.isTuple ( _arg_ )</h1>
         <p>The *isTuple* function takes one argument _arg_, and performs the following steps:</p>
         <emu-alg>
-          1. Return ? IsTuple(_arg_).
+          1. If Type(_arg_) is Tuple, return `true`.
+          1. If Type(_arg_) is Object and _arg_ has a [[TupleData]] internal slot, return `true`.
+          1. Return `false`.
         </emu-alg>
       </emu-clause>
       <emu-clause id="sec-tuple.from">
--- a/spec/immutable-data-structures.html
+++ b/spec/immutable-data-structures.html
@@ -400,8 +400,29 @@
         <p>The initial value of the *Tuple.prototype.every* is the *"every"* property of the intrinsic object %Array.prototype%.</p>
       </emu-clause>
       <emu-clause id="sec-tuple.prototype.filter">
-        <h1>Tuple.prototype.filter ()</h1>
-        <p>TODO</p>
+        <h1>Tuple.prototype.filter ( _callbackfn_ [ , _thisArg_ ] )</h1>
+        <emu-note>
+          <p>_callbackfn_ should be a function that accepts three arguments and returns a value that is coercible to the Boolean value *true* or *false*. `filter` calls _callbackfn_ once for each element in the tuple, in ascending order, and constructs a new tuple of all the values for which _callbackfn_ returns *true*.</p>
+          <p>If a _thisArg_ parameter is provided, it will be used as the *this* value for each invocation of _callbackfn_. If it is not provided, *undefined* is used instead.</p>
+          <p>_callbackfn_ is called with three arguments: the value of the element, the index of the element, and the tuple being traversed.</p>
+        </emu-note>
+        <p>When the `filter` method is called with one or two arguments, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _T_ be ? thisTupleValue(*this* value).
+          1. Let _list_ be ? CreateListFromTuple(_T_)
+          1. Let _len_ be the number of elements in _list_.
+          1. If IsCallable(_callbackfn_) is *false*, throw a *TypeError* exception.
+          1. Let _newList_ be a new empty List.
+          1. Let _k_ be 0.
+          1. Repeat, while _k_ &lt; _len_,
+            1. Let _Pk_ be ! ToString(_k_).
+            1. Let _kValue_ be _list_[_k_].
+            1. Let _selected_ be ! ToBoolean(? Call(_callbackfn_, _thisArg_, &laquo; _kValue_, _k_, _T_ &raquo;)).
+            1. If _selected_ is *true*, then
+              1. Append _kValue_ to the end of list _newList_.
+            1. Set _k_ to _k_ + 1.
+          1. Return ? CreateTupleFromList(_newList_).
+        </emu-alg>
       </emu-clause>
       <emu-clause id="sec-tuple.prototype.find">
         <h1>Tuple.prototype.find</h1>
--- a/spec/immutable-data-structures.html
+++ b/spec/immutable-data-structures.html
@@ -46,7 +46,6 @@
         <emu-alg>
           1. Perform ? RequireObjectCoercible(_iterable_).
           1. Let _fields_ be a new empty List.
-          1. Let _iteratorRecord_ be ? GetIterator(_iterable_).
           1. Let _adder_ be a new Abstract Closure with parameters (_key_, _value_) that captures _fields_ and performs the following steps when called:
             1. If Type(_value_) is Object, throw a *TypeError* exception.
             1. Let _field_ be { [[Key]]: _key_, [[Value]]: _value_ }.
@@ -122,34 +121,25 @@
         <p>When the `from` method is called with argument _items_ and optional arguments _mapfn_ and _thisArg_, the following steps are taken:</p>
         <emu-alg>
           1. If _mapfn_ is *undefined*, let _mapping_ be *false*.
-          1. Let _list_ be a new empty List.
           1. Else,
             1. If IsCallable(_mapfn_) is *false*, throw a *TypeError* exception.
             1. Let _mapping_ be *true*.
+          1. Let _list_ be a new empty List.
+          1. Let _k_ be 0.
           1. Let _usingIterator_ be ? GetMethod(_items_, @@iterator).
           1. If _usingIterator_ is not *undefined*, then
-            1. Let _iteratorRecord_ be ? GetIterator(_items_, ~sync~, _usingIterator_).
-            1. Let _k_ be 0.
-            1. Repeat,
-              1. If _k_ &ge; 2<sup>53</sup> - 1, then
-                1. Let _error_ be ThrowCompletion(a newly created *TypeError* object).
-                1. Return ? IteratorClose(_iteratorRecord_, _error_).
-              1. Let _next_ be ? IteratorStep(_iteratorRecord_).
-              1. If _next_ is *false*, then
-                1. Return a new Tuple value whose [[Sequence]] is _list_.
-              1. Let _nextValue_ be ? IteratorValue(_next_).
+            1. Let _adder_ be a new Abstract Closure with parameters (_key_, _value_) that captures (_list_, _mapFn_, _thisArg_, _mapping_, _k_) and performs the following steps when called:
               1. If _mapping_ is *true*, then
-                1. Let _mappedValue_ be Call(_mapfn_, _thisArg_, &laquo; _nextValue_, _k_ &raquo;).
-                1. If _mappedValue_ is an abrupt completion, return ? IteratorClose(_iteratorRecord_, _mappedValue_).
-                1. Set _mappedValue_ to _mappedValue_.[[Value]].
-              1. Else, let _mappedValue_ be _nextValue_.
+                1. Let _mappedValue_ be ? Call(_mapfn_, _thisArg_, &laquo; _value_, _k_ &raquo;).
+              1. Else, let _mappedValue_ be _value_.
               1. If Type(_mappedValue_) is Object, throw a *TypeError* exception.
               1. Append _mappedValue_ to _list_.
               1. Set _k_ to _k_ + 1.
+            1. Perform ! AddEntriesFromIterable(`undefined`, _iterable_, _adder_).
+            1. Return a new Tuple value whose [[Sequence]] is _list_.
           1. NOTE: _items_ is not an Iterable so assume it is an array-like object.
           1. Let _arrayLike_ be ! ToObject(_items_).
           1. Let _len_ be ? LengthOfArrayLike(_arrayLike_).
-          1. Let _k_ be 0.
           1. Repeat, while _k_ &lt; _len_,
             1. Let _Pk_ be ! ToString(_k_).
             1. Let _kValue_ be ? Get(_arrayLike_, _Pk_).
--- a/spec/immutable-data-structures.html
+++ b/spec/immutable-data-structures.html
@@ -198,7 +198,7 @@
       </emu-clause>
       <emu-clause id="sec-tuple.prototype-@@toStringTag">
         <h1>Tuple.prototype [ @@toStringTag ]</h1>
-        <p>The initial value of the *Tuple.prototype[@@toStringTag]* is the String value *"Tuple"*.</p>
+        <p>The initial value of *Tuple.prototype[@@toStringTag]* is the String value *"Tuple"*.</p>
         <p>This property has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *true* }.</p>
       </emu-clause>
       <emu-clause id="sec-tuple.prototype.popped">
@@ -419,19 +419,19 @@
       </emu-clause>
       <emu-clause id="sec-tuple.prototype.includes">
         <h1>Tuple.prototype.includes</h1>
-        <p>The initial value of the *Tuple.prototype.includes* is the *"includes"* property of the intrinsic object %Array.prototype%.</p>
+        <p>The initial value of *Tuple.prototype.includes* is %Array.prototype.includes%.</p>
       </emu-clause>
       <emu-clause id="sec-tuple.prototype.indexOf">
         <h1>Tuple.prototype.indexOf</h1>
-        <p>The initial value of the *Tuple.prototype.indexOf* is the *"indexOf"* property of the intrinsic object %Array.prototype%.</p>
+        <p>The initial value of *Tuple.prototype.indexOf* is  %Array.prototype.indexOf%.</p>
       </emu-clause>
       <emu-clause id="sec-tuple.prototype.join">
         <h1>Tuple.prototype.join</h1>
-        <p>The initial value of the *Tuple.prototype.join* is the *"join"* property of the intrinsic object %Array.prototype%.</p>
+        <p>The initial value of *Tuple.prototype.join* is  %Array.prototype.join%.</p>
       </emu-clause>
       <emu-clause id="sec-tuple.prototype.lastIndexOf">
         <h1>Tuple.prototype.lastIndexOf</h1>
-        <p>The initial value of the *Tuple.prototype.lastIndexOf* is the *"lastIndexOf"* property of the intrinsic object %Array.prototype%.</p>
+        <p>The initial value of *Tuple.prototype.lastIndexOf* is  %Array.prototype.lastIndexOf%.</p>
       </emu-clause>
       <emu-clause id="sec-tuple.prototype.entries">
         <h1>Tuple.prototype.entries ( )</h1>
@@ -444,7 +444,7 @@
       </emu-clause>
       <emu-clause id="sec-tuple.prototype.every">
         <h1>Tuple.prototype.every</h1>
-        <p>The initial value of the *Tuple.prototype.every* is the *"every"* property of the intrinsic object %Array.prototype%.</p>
+        <p>The initial value of *Tuple.prototype.every* is  %Array.prototype.every%.</p>
       </emu-clause>
       <emu-clause id="sec-tuple.prototype.filter">
         <h1>Tuple.prototype.filter ( _callbackfn_ [ , _thisArg_ ] )</h1>
@@ -472,15 +472,15 @@
       </emu-clause>
       <emu-clause id="sec-tuple.prototype.find">
         <h1>Tuple.prototype.find</h1>
-        <p>The initial value of the *Tuple.prototype.find* is the *"find"* property of the intrinsic object %Array.prototype%.</p>
+        <p>The initial value of *Tuple.prototype.find* is  %Array.prototype.find%.</p>
       </emu-clause>
       <emu-clause id="sec-tuple.prototype.findIndex">
         <h1>Tuple.prototype.findIndex</h1>
-        <p>The initial value of the *Tuple.prototype.findIndex* is the *"findIndex"* property of the intrinsic object %Array.prototype%.</p>
+        <p>The initial value of *Tuple.prototype.findIndex* is  %Array.prototype.findIndex%.</p>
       </emu-clause>
       <emu-clause id="sec-tuple.prototype.forEach">
         <h1>Tuple.prototype.forEach</h1>
-        <p>The initial value of the *Tuple.prototype.forEach* is the *"forEach"* property of the intrinsic object %Array.prototype%.</p>
+        <p>The initial value of *Tuple.prototype.forEach* is  %Array.prototype.forEach%.</p>
       </emu-clause>
       <emu-clause id="sec-tuple.prototype.keys">
         <h1>Tuple.prototype.keys ( )</h1>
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -19,7 +21,9 @@
   },
   "homepage": "https://github.com/tc39/proposal-record-tuple#readme",
   "dependencies": {
-    "ecmarkup": "^3.16.0",
+    "browser-sync": "^2.26.7",
+    "concurrently": "^5.2.0",
+    "ecmarkup": "^3.23.0",
```
rricard, line 26: can we have those changes as a separate PR so I can use them in my work? Having the spec text reload in my browser seems like a cool thing

spec/immutable-data-structures.html
```diff
@@ -6,26 +6,288 @@ <h1>Immutable Data Structures</h1>
...
+          3. Return TupleCreate(_nameList_).
+        </emu-alg>
+      </emu-clause>
+      <emu-clause id="sec-record.values">
+        <h1>Record.values ( _O_ )</h1>
+        <p>When the *values* function is called with argument _O_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _obj_ be ? ToObject(_O_)
+          2. Let _nameList_ be ? EnumerableOwnPropertyNames(_obj_, ~value~).
+          3. Return TupleCreate(_nameList_).
```
littledan, line 68: Probably the TupleCreate operation will make a wrapper object; this operation should be called something different.

rickbutton: Ah, yeah that is a good point. Something like `CreateTupleFromList`, like `CreateArrayFromList`?

cc @rricard

littledan: Sure, works for now; we can always rename later.

spec/immutable-data-structures.html
```diff
@@ -34,54 +34,41 @@ <h1>Record.isRecord ( _arg_ )</h1>
...
-        <h1>Record.values ( _O_ )</h1>
-        <p>When the *values* function is called with argument _O_, the following steps are taken:</p>
-        <emu-alg>
-          1. Let _obj_ be ? ToObject(_O_)
-          2. Let _nameList_ be ? EnumerableOwnPropertyNames(_obj_, ~value~).
-          3. Return TupleCreate(_nameList_).
-        </emu-alg>
+      <emu-clause id="sec-record.prototype">
+        <h1>Record.prototype</h1>
+        <p>The initial vlaue of *Record.prototype* is %Record.prototype%</p>
```
ljharb, line 39:
```suggestion
        <p>The initial value of *Record.prototype* is %Record.prototype%</p>

```

rricard:
> It really seems like we'll have to rename the existing concept in the spec of "Record" before landing this proposal. What are the thoughts/plans around that?

I agree, we're tracking that in #96 but maybe we need to work on that earlier because the html output is very misleading right now

spec/immutable-data-structures.html
```diff
@@ -6,26 +6,521 @@ <h1>Immutable Data Structures</h1>
...
+      </emu-clause>
+    </emu-clause>
+    <emu-clause id="sec-properties-of-the-record-constructor">
+      <h1>Properties of the Record Constructor</h1>
+      <p>The Record constructor:</p>
+      <ul>
+        <li>has a [[Prototype]] internal slot whose value is %Record.prototype%.</li>
+        <li>has the following properties:</li>
+      </ul>
+      <emu-clause id="sec-record.isrecord">
```
littledan, line 30: Record.fromEntries?

spec/immutable-data-structures.html
```diff
@@ -6,26 +6,521 @@ <h1>Immutable Data Structures</h1>
...
+      </emu-clause>
+      <emu-clause id="sec-record.prototype">
+        <h1>Record.prototype</h1>
+        <p>The initial value of *Record.prototype* is %Record.prototype%</p>
+
+        <p>This property has the attributes { [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: false }.</p>
+      </emu-clause>
+    </emu-clause>
+    <emu-clause id="sec-properties-of-the-record-prototype-object">
+      <h1>Properties of the Record Prototype Object</h1>
```
littledan, line 45: Record.prototype is null

spec/immutable-data-structures.html
```diff
@@ -6,26 +6,521 @@ <h1>Immutable Data Structures</h1>
...
+          1. Return ? CreateTupleFromList(_list_).
+        </emu-alg>
+      </emu-clause>
+      <emu-clause id="sec-tuple.prototype.values">
+        <h1>Tuple.prototype.values ()</h1>
+        <p>When the *values* method is called, it returns an iterator over the values of the tuple.</p>
+        <p>The following steps are taken:</p>
+        <emu-alg>
+          1. Let _T_ be ? thisTupleValue(*this* value).
+          1. Return CreateTupleIterator(_T_, ~value~).
```
littledan, line 506: I guess we can reuse Array iterators (here and elsewhere) as TypedArrays do.

spec/immutable-data-structures.html
```diff
@@ -6,26 +6,521 @@ <h1>Immutable Data Structures</h1>
...
+        <p>The initial value of the @@iterator property is the same function object as the initial value of the *"values"* property of the intrinsic object %Tuple.prototype%.</p>
+      </emu-clause>
+      <emu-clause id="sec-tuple.prototype.with">
+        <h1>Tuple.prototype.with ( _index_, _value_ )</h1>
+        <p>When the *with* method is called with two arguments, it returns a new tuple with the element at index _index_ replaced with value _value_.</p>
+        <emu-alg>
+          1. Let _T_ be ? thisTupleValue(*this* value).
+          2. Let _list_ be CreateListFromTuple(_T_).
+          3. Let _length_ be the length of list _list_.
+          4. If _index_ &lt; 0 or _index_ >= _length_, throw a *TypeError* exception.
```
littledan, line 520: Probably a RangeError?

spec/immutable-data-structures.html
```diff
@@ -6,26 +6,521 @@ <h1>Immutable Data Structures</h1>
...
+        <p>When the *concat* method is called with zero or more arguments, it returns a tuple containing the elements of the tuple followed by the elements of each argument in order.</p>
+        <p>The following steps are taken:</p>
+        <emu-alg>
+          1. Let _T_ be ? thisTupleValue(*this* value).
+          1. Let _list_ be a new empty List.
+          1. Let _n_ be 0.
+          1. Let _items_ be a List whose first element is _T_ and whose subsequent element are, in left to right order, the arguments that were passed to this function invocation.
+          1. Repeat, while _items_ is not empty,
+            1. Remove the first element from _items_ and let _E_ be the value of the element.
+            1. Let _spreadable_ be ? IsConcatSpreadable(_E_).
```
littledan, line 352: I guess we should make sure that Tuples are concat spreadable (either by updating that algorithm or giving them the symbol; I prefer the former).

spec/immutable-data-structures.html
rickbutton: We could totally reuse Array iterators, I just pushed a commit to add Tuple iterators, but I'm fine with either approach.

littledan: Given the precedent for TypedArrays reusing the Array iterator but making their own methods, I think you should go with Array iterators here.

rickbutton:
@littledan on `CreateListFromTuple` / `CreateTupleFromList` are you suggesting that I replace the usag e of this abstract op with the verbiage you mentioned, or define them using that verbiage? In either case, seems fine to me.

littledan:
@rickbutton I'm suggesting you replace the usages with the verbiage I mentioned.

spec/immutable-data-structures.html
```diff
@@ -275,7 +275,7 @@ <h1>Tuple.prototype.sliced ( _start_, _end_ )</h1>
         <p>The following steps are taken:</p>
         <emu-alg>
           1. Let _T_ be ? thisTupleValue(*this* value).
-          1. Let _list_ be ? CreateListFromTuple(_T_)
+          1. Let _list_ be a new List containing the elements of _T_'s [[Sequence]] value.
```
littledan, line 250: I guess in this case, you don't need a copy, and can just use T's [[Sequence]] value directly.

rickbutton: that's true, I was liberal about creating new copies of a Tuple's [[Sequence]] but I don't really need to.

spec/immutable-data-structures.html
```diff
@@ -96,7 +61,9 @@ <h1>Tuple ( ..._items_ )</h1>
           1. If NewTarget is not *undefined*, throw a *TypeError* exception.
           1. Let _elements_ be a List containing the arguments passed to this function.
           1. Let _length_ to be the number of elements in _elements_.
-          1. Let _tuple_ be ? CreateTupleFromList(_elements_)
+          1. For each element _e_ of _elements_,
+            1. If Type(_e_) is Object, throw a *TypeError* exception.
+          1. Let _tuple_ be a new Tuple value whose [[Sequence]] is _elements_.
           1. Assert: The value of _tuple_'s *"length"* property is _length_.
```
littledan, line 67: BTW this is pedantically "wrong" since primitives don't have properties. I guess you meant, Assert: ! GetV(_tuple_, *"length"*) is _length_.

rickbutton: Good point. The question really is, do we even need this check? I guess it seems superfluous, constructing a Tuple from a list will always result in a Tuple whose length is the length of said list.

littledan: I'm fine with removing the assertion too.

spec/immutable-data-structures.html
```diff
@@ -16,8 +16,20 @@ <h1>The Record Constructor</h1>
...
+        <p>When the `Record` function is called, the following steps are taken:</p>
+        <emu-alg>
+          1. If NewTarget is not *undefined*, throw a *TypeError* exception.
+          1. Let _obj_ be ? ToObject(_arg_).
+          1. Let _nameList_ be ? EnumerableOwnPropertyNames(_obj_, ~key~).
+          1. Let _fields_ be a new empty List.
+          1. For each element _name_ of _nameList_,
+            1. Let _value_ be ? GetV(_obj_, _name_).
+            1. If Type(_value_) is Object, throw a *TypeError* exception.
+            1. Let _field_ be { [[Key]]: _name_, [[Value]]: _value_ }.
```
ljharb, line 29: is it intentional that this will throw an error (observably via getters) one item at a time? If you pass `~key+value~` then all the value getting will be done up front.
```suggestion
          1. Let _nameList_ be ? EnumerableOwnPropertyNames(_obj_, ~key+value~).
          1. Let _fields_ be a new empty List.
          1. For each element _entry_ of _nameList_,
            1. Let _name_ be ! Get(_entry_, `"0"`).
            1. Let _value_ be ! Get(_entry_, `"1"`).
            1. If Type(_value_) is Object, throw a *TypeError* exception.
            1. Let _field_ be { [[Key]]: _name_, [[Value]]: _value_ }.
```

spec/immutable-data-structures.html
```diff
@@ -27,6 +39,32 @@ <h1>Properties of the Record Constructor</h1>
...
+            1. Let _nextItem_ be ? IteratorValue(_next_).
+            1. If Type(_nextItem_) is not Object, then
+              1. Let _error_ be ThrowCompletion(a newly created *TypeError* object).
+              1. Return ? IteratorClose(_iteratorRecord_, _error_).
+            1. Let _k_ be Get(_nextItem_, *"0"*).
+            1. If _k_ is an abrupt completion, return ? IteratorClose(_iteratorRecord_, _k_).
+            1. Let _v_ be Get(_nextItem_, *"1"*).
+            1. If _v_ is an abrupt completion, return ? IteratorClose(_iteratorRecord_, _v_).
+            1. Let _field_ be { [[Key]]: _k_, [[Value]]: _v_ }.
+            1. Append _field_ to the end of list _fields_.
```
ljharb, line 62: This seems like it could benefit from using an abstract operation like AddEntriesFromIterable, where the `adder` creates a List, and then at the end, return the new Record?

spec/immutable-data-structures.html
rickbutton: It was intentional, although an argument could be made either way. @littledan suggested that I check/throw in each iteration of `Tuple.prototype.map`, and I figured that logic should apply here as well.

In general, what is the approach for things like this? Is there prior art or reasoning for interleaving the "gets" and type checks vs saving type checks for the end?

spec/immutable-data-structures.html
rickbutton: I based this originally on `AddEntriesFromIterable` in the first place, so that is fine by me. I initially simplified this to remove an additional abstract op since the resulting op would initially be only used here. I guess something like `CreateRecordFieldsFromIterable` (as you said, returning a list of fields)?

spec/immutable-data-structures.html
ljharb: I don't know if there's a strong precedent either way; when using an iterable, throwing midway seems preferable, but for objects, i could see it going either way. To me it doesn't make much sense to throw midway through iterating an object, since an object isn't an iterable, it's a "bag of stuff".

spec/immutable-data-structures.html
ljharb: The main difference is the "checking if it's an object and throwing". However if the `adder` did this check before appending to a list, it seems like it would work?

spec/immutable-data-structures.html
rickbutton: Seems reasonable to me. I switched to using `key+value`.

spec/immutable-data-structures.html
```diff
@@ -21,10 +21,11 @@ <h1>Record ( _arg_ )</h1>
...
           1. If NewTarget is not *undefined*, throw a *TypeError* exception.
           1. Let _obj_ be ? ToObject(_arg_).
-          1. Let _nameList_ be ? EnumerableOwnPropertyNames(_obj_, ~key~).
+          1. Let _props_ be ? EnumerableOwnPropertyNames(_obj_, ~key+value~).
           1. Let _fields_ be a new empty List.
-          1. For each element _name_ of _nameList_,
-            1. Let _value_ be ? GetV(_obj_, _name_).
+          1. For each element _prop_ of _props_,
+            1. Let _name_ be ? GetV(_obj_, `"0"`).
+            1. Let _value_ be ? GetV(_obj_, `"1"`).
```
ljharb, line 28:
```suggestion
            1. Let _name_ be ! GetV(_prop_, `"0"`).
            1. Let _value_ be ! GetV(_prop_, `"1"`).
```
(the `!` is since we know these can't fail, because of `EnumerableOwnPropertyNames`'s algorithm)

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+        <emu-alg>
+          1. If NewTarget is not *undefined*, throw a *TypeError* exception.
+          1. Let _obj_ be ? ToObject(_arg_).
+          1. Let _props_ be ? EnumerableOwnPropertyNames(_obj_, ~key+value~).
+          1. Let _fields_ be a new empty List.
+          1. For each element _prop_ of _props_,
+            1. Let _name_ be ! GetV(_prop_, `"0"`).
+            1. Let _value_ be ! GetV(_prop_, `"1"`).
+            1. If Type(_value_) is Object, throw a *TypeError* exception.
+            1. Let _field_ be { [[Key]]: _name_, [[Value]]: _value_ }.
```
ljharb, line 30:
```suggestion
            1. Let _field_ be the Record { [[Key]]: _name_, [[Value]]: _value_ }.

```
altho this would have to use the new name for spec-records :-)

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+        <emu-note>
+          <p>The parameter _iterable_ is expected to be an object that implements an @@iterator method that returns an iterator object that produces a two element array-like object whose first element is a value that will be used as a Map key and whose second element is the value to associate with that key.</p>
+        </emu-note>
+      </emu-clause>
+      <emu-clause id="sec-append-record-field-to-list-functions">
+        <h1>AppendRecordFieldToList Functions</h1>
+        <p>An AppendRecordFieldToList function is an anonymous built-in function. When an AppendRecordFieldtoList function is called with arguments _key_ and _value_, the following steps are taken:</p>
+        <emu-alg>
+            1. Let _list_ be the *this* value.
+            1. Assert: Type(_list_) is List.
```
ljharb, line 64: List isn't a JS type, and Type() can never return "List".

I'm not sure what to do here; the receiver of a function (even a spec artifice) has to be a JS value.

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+            1. If Type(_value_) is Object, throw a *TypeError* exception.
+            1. Let _field_ be { [[Key]]: _key_, [[Value]]: _value_ }.
+            1. Append _field_ to the end of list _list_.
+        </emu-alg>
+      </emu-clause>
+      <emu-clause id="sec-record.isrecord">
+        <h1>Record.isRecord ( _arg_ )</h1>
+        <p>The *isRecord* function takes one argument _arg_, and performs the following steps:</p>
+        <emu-alg>
+          1. Return ? IsRecord(_arg_).
```
ljharb, line 74: Where is this defined, and when could it return an abrupt completion?

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+      <emu-clause id="sec-tuple-items">
+        <h1>Tuple ( ..._items_ )</h1>
+        <p>When the `Tuple` function is called, the following steps are taken:</p>
+        <emu-alg>
+          1. If NewTarget is not *undefined*, throw a *TypeError* exception.
+          1. Let _elements_ be a List containing the arguments passed to this function.
+          1. Let _length_ to be the number of elements in _elements_.
+          1. For each element _e_ of _elements_,
+            1. If Type(_e_) is Object, throw a *TypeError* exception.
+          1. Let _tuple_ be a new Tuple value whose [[Sequence]] is _elements_.
```
ljharb, line 106: to use the common wording in the spec:
```suggestion
          1. Let _items_ be the list of arguments passed to this function.
          1. Let _length_ to be the number of elements in _items_.
          1. For each element _e_ of _items_,
            1. If Type(_e_) is Object, throw a *TypeError* exception.
          1. Let _tuple_ be a new Tuple value whose [[Sequence]] is _items_.
```

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+      <p>The Tuple constructor:</p>
+      <ul>
+        <li>has a [[Prototype]] internal slot whose value is %Tuple.prototype%.</li>
+        <li>has the following properties:</li>
+      </ul>
+      <emu-clause id="sec-tuple.istuple">
+        <h1>Tuple.isTuple ( _arg_ )</h1>
+        <p>The *isTuple* function takes one argument _arg_, and performs the following steps:</p>
+        <emu-alg>
+          1. Return ? IsTuple(_arg_).
```
ljharb, line 122: same question here re IsTuple - where is it defined, and when could it return abruptly?

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+      <emu-clause id="sec-tuple.from">
+        <h1>Tuple.from ( _items_ [ , _mapFn_ [, _thisArg_ ] ] )</h1>
+        <p>When the `from` method is called with argument _items_ and optional arguments _mapfn_ and _thisArg_, the following steps are taken:</p>
+        <emu-alg>
+          1. If _mapfn_ is *undefined*, let _mapping_ be *false*.
+          1. Let _list_ be a new empty List.
+          1. Else,
+            1. If IsCallable(_mapfn_) is *false*, throw a *TypeError* exception.
+            1. Let _mapping_ be *true*.
+          1. Let _usingIterator_ be ? GetMethod(_items_, @@iterator).
```
ljharb, line 134: could this whole function use AddEntriesFromIterable instead of duplicating Array.from?

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+      </emu-clause>
+      <emu-clause id="sec-tuple.of">
+        <h1>Tuple.of ( ..._items_ )</h1>
+        <p>The *of* method takes any number of arguments, and performs the following steps:</p>
+        <emu-alg>
+          1. Let _elements_ be a List containing the arguments passed to this function.
+          1. Let _length_ to be the number of elements in _elements_.
+          1. For each element _e_ of _elements_,
+            1. If Type(_e_) is Object, throw a *TypeError* exception.
+          1. Let _tuple_ be a new Tuple value whose [[Sequence]] is _elements_.
```
ljharb, line 178:
```suggestion
          1. Let _items_ be the List of arguments passed to this function.
          1. Let _length_ to be the number of elements in _items_.
          1. For each element _e_ of _items_,
            1. If Type(_e_) is Object, throw a *TypeError* exception.
          1. Let _tuple_ be a new Tuple value whose [[Sequence]] is _items_.

```

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+          1. Return _tuple_.
+        </emu-alg>
+      </emu-clause>
+      <emu-clause id="sec-tuple.prototype">
+        <h1>Tuple.prototype</h1>
+        <p>The initial vlaue of *Tuple.prototype* is %Tuple.prototype%</p>
```
ljharb, line 184:
```suggestion
```
dotted intrinsics don't need this line

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+        <p>This property has the attributes { [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: false }.</p>
       </emu-clause>
     </emu-clause>
+    <emu-clause id="sec-properties-of-the-tuple-prototype-object">
+      <h1>Properties of the Tuple Prototype Object</h1>
+      <p>The Tuple prototype object:</p>
+      <ul>
+        <li>is an ordinary object.</li>
+        <li>is not a Tuple object; it does not have a [[TupleData]] internal slot.</li>
+        <li>has a [[Prototype]] internal slot whose value is the value *null*.</li>
```
ljharb, line 194:
```suggestion
        <li>has a [[Prototype]] internal slot whose value is *null*.</li>

```

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+        </emu-alg>
+      </emu-clause>
+      <emu-clause id="sec-tuple.prototype-@@toStringTag">
+        <h1>Tuple.prototype [ @@toStringTag ]</h1>
+        <p>The initial value of the *Tuple.prototype[@@toStringTag]* is the String value *"Tuple"*.</p>
+        <p>This property has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *true* }.</p>
+      </emu-clause>
+      <emu-clause id="sec-tuple.prototype.popped">
+        <h1>Tuple.prototype.popped ()</h1>
+        <p>When the *popped* method is called, it returns a tuple containing the elements of the tuple, except for the last value.</p>
```
ljharb, line 223:
```suggestion
        <p>When the *popped* method is called, it returns a Tuple containing the elements of the Tuple, except for the last value.</p>

```

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+          1. Let _length_ be the number of elements in _list_.
+          1. If _length_ is 0,
+            1. Return _T_.
+          1. Remove the last element of _list_.
+          1. Return a new Tuple value whose [[Sequence]] is _list_.
+        </emu-alg>
+      </emu-clause>
+      <emu-clause id="sec-tuple.prototype.pushed">
+        <h1>Tuple.prototype.pushed ( ..._args_ )</h1>
+        <p>When the *pushed* method is called, it returns a tuple containing the elements of the tuple, followed by the arguments in order.</p>
```
ljharb, line 237:
```suggestion
        <p>When the *pushed* method is called, it returns a Tuple containing the elements of the Tuple, followed by the arguments in order.</p>

```

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+        <emu-alg>
+          1. Let _T_ be ? thisTupleValue(*this* value).
+          1. Let _list_ be a new List containing the elements of _T_'s [[Sequence]] value.
+          1. Let _argsList_ be a List containing the arguments passed to this function invocation.
+          1. Let _n_ be the number of elements in _list_.
+          1. Repeat, while _argsList_ is not empty,
+            1. Remove the first element from _argsList_, and let _E_ be the value of the element.
+            1. If _n_ + 1 &gt; 2<sup>53</sup> - 1, throw a *TypeError* exception.
+            1. If Type(_E_) is Object, throw a *TypeError* exception.
+            1. Append _E_ to the end of list _list_.
```
ljharb, line 248:
```suggestion
            1. Append _E_ to the end of List _list_.
```

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+        <h1>Tuple.prototype.reversed ()</h1>
+        <p>When the *reversed* method is called, it returns a tuple containing the elements of the tuple, in reverse order.</p>
+        <p>The following steps are taken:</p>
+        <emu-alg>
+          1. Let _T_ be ? thisTupleValue(*this* value).
+          1. Let _oldList_ be a new List containing the elements of _T_'s [[Sequence]] value.
+          1. Let _newList_ be a new empty List.
+          1. Repeat, while _oldList_ is not empty,
+            1. Remove the first element from _oldList_, and let _E_ be the value of the element.
+            1. Append _E_ to the end of list _newList_.
```
ljharb, line 263:
```suggestion
            1. Append _E_ to the end of List _newList_.
```

spec/immutable-data-structures.html
ljharb: It's a protocol; it's fine to have default behavior here, but I think the symbol needs to exist and have its value respected.

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+        <h1>Tuple.prototype.includes</h1>
+        <p>The initial value of the *Tuple.prototype.includes* is the *"includes"* property of the intrinsic object %Array.prototype%.</p>
+      </emu-clause>
+      <emu-clause id="sec-tuple.prototype.indexOf">
+        <h1>Tuple.prototype.indexOf</h1>
+        <p>The initial value of the *Tuple.prototype.indexOf* is the *"indexOf"* property of the intrinsic object %Array.prototype%.</p>
+      </emu-clause>
+      <emu-clause id="sec-tuple.prototype.join">
+        <h1>Tuple.prototype.join</h1>
+        <p>The initial value of the *Tuple.prototype.join* is the *"join"* property of the intrinsic object %Array.prototype%.</p>
```
ljharb, line 447:
```suggestion
        <p>The initial value of *Tuple.prototype.join* is %Array.prototype.join%.</p>
```

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+          1. Return a new Tuple value whose [[Sequence]] is _list_.
+        </emu-alg>
+      </emu-clause>
+      <emu-clause id="sec-tuple.prototype.includes">
+        <h1>Tuple.prototype.includes</h1>
+        <p>The initial value of the *Tuple.prototype.includes* is the *"includes"* property of the intrinsic object %Array.prototype%.</p>
+      </emu-clause>
+      <emu-clause id="sec-tuple.prototype.indexOf">
+        <h1>Tuple.prototype.indexOf</h1>
+        <p>The initial value of the *Tuple.prototype.indexOf* is the *"indexOf"* property of the intrinsic object %Array.prototype%.</p>
```
ljharb, line 443:
```suggestion
        <p>The initial value of *Tuple.prototype.indexOf* is  %Array.prototype.indexOf%.</p>
```

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+              1. If _n_ &ge; 2<sup>53</sup> - 1, throw a *TypeError* exception.
+              1. If Type(_E_) is Object, throw a *TypeError* exception.
+              1. Append _E_ to the end of list _list_.
+              1. Set _n_ to _n_ + 1.
+          1. Return a new Tuple value whose [[Sequence]] is _list_.
+        </emu-alg>
+      </emu-clause>
+      <emu-clause id="sec-tuple.prototype.includes">
+        <h1>Tuple.prototype.includes</h1>
+        <p>The initial value of the *Tuple.prototype.includes* is the *"includes"* property of the intrinsic object %Array.prototype%.</p>
```
ljharb, line 439:
```suggestion
        <p>The initial value of *Tuple.prototype.includes* is %Array.prototype.includes%.</p>
```

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+        <h1>Tuple.prototype.indexOf</h1>
+        <p>The initial value of the *Tuple.prototype.indexOf* is the *"indexOf"* property of the intrinsic object %Array.prototype%.</p>
+      </emu-clause>
+      <emu-clause id="sec-tuple.prototype.join">
+        <h1>Tuple.prototype.join</h1>
+        <p>The initial value of the *Tuple.prototype.join* is the *"join"* property of the intrinsic object %Array.prototype%.</p>
+      </emu-clause>
+      <emu-clause id="sec-tuple.prototype.lastIndexOf">
+        <h1>Tuple.prototype.lastIndexOf</h1>
+        <p>The initial value of the *Tuple.prototype.lastIndexOf* is the *"lastIndexOf"* property of the intrinsic object %Array.prototype%.</p>
```
ljharb, line 451:
```suggestion
        <p>The initial value of *Tuple.prototype.lastIndexOf* is %Array.prototype.lastIndexOf%.</p>
```

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+        <p>When the *entries* method is called, it returns an iterator over the entries of the tuple.</p>
+        <p>The following steps are taken:</p>
+        <emu-alg>
+          1. Let _O_ be ? ToObject(*this* value).
+          1. Return CreateArrayIterator(_O_, ~key+value~).
+        </emu-alg>
+      </emu-clause>
+      <emu-clause id="sec-tuple.prototype.every">
+        <h1>Tuple.prototype.every</h1>
+        <p>The initial value of the *Tuple.prototype.every* is the *"every"* property of the intrinsic object %Array.prototype%.</p>
```
ljharb, line 464:
```suggestion
        <p>The initial value of the *Tuple.prototype.every* is %Array.prototype.every%.</p>
```

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+            1. Let _selected_ be ! ToBoolean(? Call(_callbackfn_, _thisArg_, &laquo; _kValue_, _k_, _T_ &raquo;)).
+            1. If _selected_ is *true*, then
+              1. Append _kValue_ to the end of list _newList_.
+            1. Set _k_ to _k_ + 1.
+          1. Return a new Tuple value whose [[Sequence]] is _newList_.
+        </emu-alg>
+      </emu-clause>
+      <emu-clause id="sec-tuple.prototype.find">
+        <h1>Tuple.prototype.find</h1>
+        <p>The initial value of the *Tuple.prototype.find* is the *"find"* property of the intrinsic object %Array.prototype%.</p>
```
ljharb, line 492:
```suggestion
        <p>The initial value of *Tuple.prototype.find* is %Array.prototype.find%.</p>
```

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+      </emu-clause>
+      <emu-clause id="sec-tuple.prototype.findIndex">
+        <h1>Tuple.prototype.findIndex</h1>
+        <p>The initial value of the *Tuple.prototype.findIndex* is the *"findIndex"* property of the intrinsic object %Array.prototype%.</p>
```
ljharb, line 496:
```suggestion
        <p>The initial value of *Tuple.prototype.findIndex* is %Array.prototype.findIndex%.</p>
```

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+      </emu-clause>
+      <emu-clause id="sec-tuple.prototype.forEach">
+        <h1>Tuple.prototype.forEach</h1>
+        <p>The initial value of the *Tuple.prototype.forEach* is the *"forEach"* property of the intrinsic object %Array.prototype%.</p>
```
ljharb, line 500:
```suggestion
        <p>The initial value of *Tuple.prototype.forEach* is %Array.prototype.forEach%.</p>
```

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+            1. Set _k_ to _k_ + 1.
+          1. Return a new Tuple value whose [[Sequence]] is _newList_.
+        </emu-alg>
+      </emu-clause>
+      <emu-clause id="sec-tuple.prototype.reduceRight">
+        <h1>Tuple.prototype.reduceRight ()</h1>
+      </emu-clause>
+      <emu-clause id="sec-tuple.prototype.some">
+        <h1>Tuple.prototype.some</h1>
+        <p>The initial value of the *Tuple.prototype.some* is the *"some"* property of the intrinsic object %Array.prototype%.</p>
```
ljharb, line 540:
```suggestion
        <p>The initial value of the *Tuple.prototype.some* is %Array.prototype.some%.</p>

```

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,590 @@ <h1>Immutable Data Structures</h1>
...
+        <p>When the *values* method is called, it returns an iterator over the values of the tuple.</p>
+        <p>The following steps are taken:</p>
+        <emu-alg>
+          1. Let _O_ be ? ToObject(*this* value).
+          1. Return CreateArrayIterator(_O_, ~value~).
+        </emu-alg>
+      </emu-clause>
+      <emu-clause id="sec-tuple.prototype-@@iterator">
+        <h1>Tuple.prototype [ @@iterator ] ()</h1>
+        <p>The initial value of the @@iterator property is the same function object as the initial value of the *"values"* property of the intrinsic object %Tuple.prototype%.</p>
```
ljharb, line 577:
```suggestion
        <p>The initial value of the @@iterator property is %Tuple.prototype.values%.</p>

```

spec/immutable-data-structures.html
littledan: Agree that the receiver can't be a spec List, but there should be a ton of ways to fix this depending on what you're trying to do. I'm not sure what to do either, though, as I don't see any callers of this function.

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,596 @@ <h1>Immutable Data Structures</h1>
...
+            1. Append _field_ to the end of list _fields_.
+          1. Return a new Record value whose [[Fields]] is _fields_.
+        </emu-alg>
+      </emu-clause>
+    </emu-clause>
+    <emu-clause id="sec-properties-of-the-record-constructor">
+      <h1>Properties of the Record Constructor</h1>
+      <p>The Record constructor:</p>
+      <ul>
+        <li>has a [[Prototype]] internal slot whose value is %Record.prototype%.</li>
```
littledan, line 41: The value is null

spec/immutable-data-structures.html
littledan: Oops, I see the cross-reference is by id

littledan: Rather than threading the fields through the `this` value, I suggest using *undefined* for the `this` value, and use an [abstract closure](https://tc39.es/ecma262/#sec-abstract-closure) to capture the _fields_ list.

spec/immutable-data-structures.html
littledan: Probably this could be an ! ; note that there are multiple parallel PRs in development, so another one which covers the data model could backfill the definition of IsRecord.

Early reviews can be good, but let's focus on offering constructive suggestions to newer spec authors. There will be lots of time to pick things apart for remaining errors after the champion group and collaborators collectively write this text.

At the same time, I think we're far along enough in the development of the data model that we can fill in concrete text here. I think the `IsRecord` algorithm will not have any other uses, so we can describe this concretely. It could look something like this:

1. If Type(_arg_) is Record, return *true*.
1. If Type(_arg_) is Object and _arg_ has a [[RecordData]] internal slot, return *true*.
1. Return *false*.

Similar wording could be used for Tuple.isTuple.

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,596 @@ <h1>Immutable Data Structures</h1>
...
-        <h1>Record ( _value_ )</h1>
-        <p>TBD</p>
+        <h1>Record ( _arg_ )</h1>
+        <p>When the `Record` function is called, the following steps are taken:</p>
+        <emu-alg>
+          1. If NewTarget is not *undefined*, throw a *TypeError* exception.
+          1. Let _obj_ be ? ToObject(_arg_).
+          1. Let _props_ be ? EnumerableOwnPropertyNames(_obj_, ~key+value~).
+          1. Let _fields_ be a new empty List.
+          1. For each element _prop_ of _props_,
```
littledan, line 26: You can't iterate over iterators like this; "For each" is for spec lists, not JS values. It'd be easier to iterate over the property keys by copy-pasting the first 4 steps of EnumerableOwnPropertyNames here.

spec/immutable-data-structures.html
littledan: I'm not sure what you're trying to say here. Nothing in the JS spec has that symbol defined. I'm not saying that we would stop querying the symbol, just that it would have this default behavior built into the IsConcatSpreadable algorithm, like Arrays do.

spec/immutable-data-structures.html
ljharb: I find all of my suggestions constructive; you’re free to apply your own rubric to your own suggestions.

rickbutton: To clarify, I am really thankful for all of the comments from @ljharb and @littledan, especially as a brand-new spec author. I'm sure there is a massive list of work left to do here, and am eager to soak up as much info as possible.

package.json
rickbutton: @rricard sorry for the delay on this, I put these package.json changes on master, so if you rebase you should have them now.

spec/immutable-data-structures.html
rickbutton: At least the first half can, since the second half handles array-likes.

spec/immutable-data-structures.html
rickbutton: @littledan doesn't `EnumerableOwnPropertyNames` return a spec-List, not an iterable?

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,575 @@ <h1>Immutable Data Structures</h1>
...
+        <li>creates and initializes a new Tuple object when called as a function.</li>
+        <li>is not intended to be used with the `new` operator or to be subclassed. It may be used as the value of an *extends* clause of a class definition but a *super* call to the Tuple constructor will cause an exception.</li>
+      </ul>
+      <emu-clause id="sec-tuple-items">
+        <h1>Tuple ( ..._items_ )</h1>
+        <p>When the `Tuple` function is called, the following steps are taken:</p>
+        <emu-alg>
+          1. If NewTarget is not *undefined*, throw a *TypeError* exception.
+          1. Let _elements_ be a List containing the arguments passed to this function.
+          1. Let _length_ to be the number of elements in _elements_.
```
ljharb, line 95:
```suggestion
```

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,575 @@ <h1>Immutable Data Structures</h1>
...
+            1. Set _k_ to _k_ + 1.
+          1. Return a new Tuple value whose [[Sequence]] is _list_.
+        </emu-alg>
+      </emu-clause>
+      <emu-clause id="sec-tuple.of">
+        <h1>Tuple.of ( ..._items_ )</h1>
+        <p>The *of* method takes any number of arguments, and performs the following steps:</p>
+        <emu-alg>
+          1. Let _elements_ be a List containing the arguments passed to this function.
+          1. Let _length_ to be the number of elements in _elements_.
```
ljharb, line 160:
```suggestion
```

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,575 @@ <h1>Immutable Data Structures</h1>
...
+      </emu-alg>
+      <emu-clause id="sec-tuple.prototype.constructor">
+        <h1>Tuple.prototype.constructor</h1>
+        <p>The initial value of *Tuple.prototype.constructor* is the intrinsic object %Tuple%</p>
+      </emu-clause>
+      <emu-clause id="sec-tuple.prototype.valueof">
+        <h1>Tuple.prototype.valueOf ( )</h1>
+        <p>When the `valueOf` function is called, the following steps are taken:</p>
+        <emu-alg>
+          1. Return ? thisTupleValue(*this* value)
```
ljharb, line 198:
```suggestion
          1. Return ? thisTupleValue(*this* value).
```

spec/immutable-data-structures.html
ljharb: this one still isn't resolved?

spec/immutable-data-structures.html
ljharb: this one still isn't resolved?

spec/immutable-data-structures.html
ljharb: also not yet resolved.

spec/immutable-data-structures.html
rickbutton: Yeah somehow these committed suggestions got lost. I just fixed all of the occurrences of this again.

rickbutton:
Oh, I realize what happened @ljharb , I messed up a force-push. Whoops! I will reapply your suggestions asap.

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,573 @@ <h1>Immutable Data Structures</h1>
...
-        <p>TBD</p>
+      <p>The Tuple constructor:</p>
+      <ul>
+        <li>is the intrinsic object <dfn>%Tuple%</dfn>.</li>
+        <li>is the initial value of the *"Tuple"* property of the global object.</li>
+        <li>creates and initializes a new Tuple object when called as a function.</li>
+        <li>is not intended to be used with the `new` operator or to be subclassed. It may be used as the value of an *extends* clause of a class definition but a *super* call to the Tuple constructor will cause an exception.</li>
+      </ul>
+      <emu-clause id="sec-tuple-items">
+        <h1>Tuple ( ..._items_ )</h1>
```
littledan, line 90: Hmm, do we want this to work, or do we want to require that people use Tuple.of for this functionality? They're sort of duplicated, and this isn't parallel to Array.

spec/immutable-data-structures.html
```diff
@@ -6,27 +6,573 @@ <h1>Immutable Data Structures</h1>
...
+      <emu-clause id="sec-array.prototype.sort">
+        <h1>Tuple.prototype.sorted ( _comparefn_ )</h1>
+        <p>A new tuple the elements from the original tuple is returned, where the elements are sorted. The sort must be stable (that is, elements that compare equal must remain in their original order). If _comparefn_ is not *undefined*, it should be a function that accepts two arguments _x_ and _y_ and returns a negative value if _x_ &lt; _y_, zero if _x_ = _y_, or a positive value if _x_ &gt; _y_.</p>
+        <p>Upon entry, the following steps are performed to initialize evaluation of the `sorted` function:</p>
+        <emu-alg>
+          1. If _comparefn_ is not *undefined* and IsCallable(_comparefn_) is *false*, throw a *TypeError* exception.
+          1. Let _T_ be ? thisTupleValue(*this* value).
+          1. Let _list_ be a new List containing the elements of _T_'s [[Sequence]] value.
+          1. Let _len_ be the number of elements in _list_.
+        </emu-alg>
```
littledan, line 298: It would be good if you could avoid duplicating the below section with Array.prototype.sort, since it's just the same, right?

spec/immutable-data-structures.html
littledan: Why was this change made? The previous wording seems simpler.

spec/immutable-data-structures.html
rickbutton: apologies, part of my broken push, I've pushed a commit that contains all of @ljharb 's suggested changes. Should be good now.

spec/immutable-data-structures.html
rickbutton: Most of it is the same, except with the handling of sparse arrays removed. We could share more here. What would be the best way to share more of this? Not sure how to handle it, since it isn't quite an abstract-op, very prose-y.

spec/immutable-data-structures.html
rickbutton: For context, we had an out of band discussion about this and figured that it would make sense to, at least for now, do the same thing as `Array` here, where we create a `Tuple` full of `n` `undefined` values when called as `Tuple(n)`, the same single-argument-overload as `Array`.

rickbutton:
Merging this in so that we can continue to iterate on this in smaller chunks and merge it into the other and upcoming PRs.

rricard:
I'll rebase on this as well since you improved the ecmarkup pipeline

PR #132, opened 2020-06-10
Record and Tuple object model spec

rricard:


spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,90 @@
...
+          1. Return ! RecordGetOwnProperty(_R_, _P_).
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-record-exotic-objects-defineownproperty-p-desc">
+        <h1>[[DefineOwnProperty]] ( _P_, _Desc_ )</h1>
+        <p>When the [[DefineOwnProperty]] internal method of a Record exotic object _R_ is called with property key _P_, and Property Descriptor _Desc_ the following steps are taken:</p>
+        <emu-alg>
+          1. Assert: IsPropertyKey(_P_) is *true*.
+          1. Throw a *TypeError* exception.
```
littledan, line 33: This needs to accept property definitions that don't change the value. See module namespace exotic objects for an example. The same goes for setting the prototype. https://tc39.es/ecma262/#sec-module-namespace-exotic-objects

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,90 @@
...
+            1. Assert Type(rec) is Record
+            1. Let _len_ be the length of _rec_.
+            1. Let _resultPrim_ be *undefined*
+            1. For each primitive data property _pdp_, do
+              1. Let _key_ be the String key value of _pdp_
+              1. Let _prim_ be the associated value of _pdp_
+              1. Assert: Type(_key_) is String
+              1. Assert: Type(_prim_) is not Object
+              1. If _key_ == _P_, let _resultPrim_ be _prim_
+            1. If _resultPrim_ is *undefined*, return *undefined*
```
littledan, line 84: It should be possible to have a data property of a Record whose value is *undefined*, so this seems a little broken. Maybe you could start _resultPrim_ as ~empty~ and check for that here instead.

spec/data-types-and-values.html
```diff
@@ -7,7 +7,10 @@ <h1>ECMAScript Language Types</h1>
 
     <emu-clause id="sec-ecmascript-language-types-record-type">
       <h1>The Record Type</h1>
-      <p>TBD</p>
+      <p>The Record type is a coumpound primitive type. It is an ordered collection of primitive data properties.</p>
+      <p>A <dfn>primitive data property</dfn> associates a String key value with a primitive value or a coumpound primitive value called associated value.</p>
```
littledan, line 11: I can see how you've derived this from the spec text for objects. I want to suggest that you take a more explicit approach, with Lists and Records to represent the contents, for a couple reasons:
- Juggling the notions of "the Record has a primitive data property" from "the Record wrapper has a data property" feels complicated. I'd prefer to think about just one property (the "virtual" one in the wrapper)
- This vague text raises more questions than it answers, since it it ambiguous about whether property attributes, ordering are supported.
- Newer constructs (like private fields and methods) end up having representations in terms of Records and Lists, and I'd like to continue this with newer things.

I would instead say, Record values have an associated List of Records of the form { [[Key]], [[Value]] }, where [[Key]] is a String and [[Value]] is any primitive (including compound primitives), which are sorted by [[Key]] in code unit order. Operations which create Record values are responsible for setting this up appropriately. The List of Records is never modified.

This PR does not contain any operation which creates new Record values. Maybe you could start with the definition of Record literals. Without an example like that, I can't really review how this definition works in practice.

rricard:
Thanks dan, I'll make another pass on this tomorrow

spec/data-types-and-values.html
rricard: > This PR does not contain any operation which creates new Record values. Maybe you could start with the definition of Record literals. Without an example like that, I can't really review how this definition works in practice.

I will do that, in the meantime, in which section do you think this should appear?

littledan: It's fine to put it here. We could consider moving it around when merging into the main spec.

spec/data-types-and-values.html
```diff
@@ -13,6 +13,21 @@ <h1>The Record Type</h1>
       <p>
         The Record type is used to represent the association of multiple primitive values together. Each pair is regarded as occupying a position within the sequence. These positions are indexed with nonnegative integers. The first pair (if any) is at index 0, the next element (if any) at index 1, and so on. The length of a Record is the number of pairs within it. The empty Record has length zero and therefore contains no elements.
       </p>
+
+      <emu-clause id="sec-recordtypecreate" aoid="RecordTypeCreate">
+        <h1>RecordTypeCreate ( _keysList_, _valuesList_ )</h1>
+        <p>This abstract operation RecordTypeCreate takes a _keysList_ and a _valuesList_ argument. It is used to specify the creation of new Record value types. It performs the following steps when called:</p>
+        <emu-alg>
+            1. Assert: Type(_keysList_) is a List of Strings.
+            1. Assert: Type(_valuesList_) is a List of primitive values.
```
ljharb, line 22: `Type()` is for JS values, a List isn't one.
```suggestion
            1. Assert: _keysList_ is a List of Strings.
            1. Assert: _valuesList_ is a List of primitive values.

```

spec/data-types-and-values.html
```diff
@@ -13,6 +13,21 @@ <h1>The Record Type</h1>
...
+        <h1>RecordTypeCreate ( _keysList_, _valuesList_ )</h1>
+        <p>This abstract operation RecordTypeCreate takes a _keysList_ and a _valuesList_ argument. It is used to specify the creation of new Record value types. It performs the following steps when called:</p>
+        <emu-alg>
+            1. Assert: Type(_keysList_) is a List of Strings.
+            1. Assert: Type(_valuesList_) is a List of primitive values.
+            1. Assert: _valuesList_.length == _keysList_.length
+            1. Let _L_ be an empty List.
+            1. For each _i_ from 0 to _keysList_.length:
+              1. Add { [[Key]]: _keysList_[_i_], [[Value]]: _valuesList_[_i_] } to _L_
+            1. Sort _L_ by [[Key]] using code unit order
```
ljharb, line 27: this seems like it should be fleshed out more precisely

rricard: Agreed I took a shortcut here, I will definitely come back on it later

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,97 @@
...
+          1. For each primitive data property _pdp_, do
+            1. Let _key_ be the String key value of _pdp_
+            1. Assert: Type(_key_) is String
+            1. Add _key_ to _keys_
+          1. Return _keys_
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-recordcreate" aoid="RecordCreate">
+        <h1>RecordCreate ( _value_ )</h1>
```
littledan, line 61: Look at [Module namespace exotic objects](https://tc39.es/ecma262/#module-namespace-exotic-object) to see how to specify an exotic object that can't be modified. (Note that some things will be different, e.g., those show writable true, but we show writeable false. But module namespace exotic objects override all the methods you have to override.)

spec/data-types-and-values.html
```diff
@@ -7,12 +7,35 @@ <h1>ECMAScript Language Types</h1>
...
+            1. Return _L_
+        </emu-alg>
+      </emu-clause>
     </emu-clause>
 
     <emu-clause id="sec-ecmascript-language-types-tuple-type">
       <h1>The Tuple Type</h1>
-      <p>TBD</p>
+      <p>
```
littledan, line 36: At some point, you'll want to look at [Integer Indexed exotic objects](https://tc39.es/ecma262/#sec-integer-indexed-exotic-objects) to inform how property access works here.

spec/overview.html
```diff
@@ -47,8 +47,7 @@ <h1>Terms and Definitions</h1>
     <emu-clause id="sec-record-value">
       <h1>Record value</h1>
       <p>
-        primitive value that is a structure mapping a set of string values to
-        any other primitive values
+        primitive value that is a sorted list of data. Each item of the list has a String value key and a primitive value and the list is itself sorted by key.
```
littledan, line 50: Following my proposed wording above, this could be:

> primitive value which is a finite mapping from Strings to ECMAScript primitive values.

spec/data-types-and-values.html
ljharb: "using code unit order" doesn't describe an algorithm to me ¯\\\_(ツ)_/¯

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,97 @@
...
+
+      <emu-clause id="sec-record-exotic-objects-ownpropertykeys">
+        <h1>[[OwnPropertyKeys]] ( )</h1>
+        <p>When the [[OwnPropertyKeys]] internal method of a Record exotic object _R_ is called, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _keys_ be a new empty List.
+          1. Let _rec_ be _R_.[[RecordData]].
+          1. Assert: Type(_rec_) is Record.
+          1. Let _len_ be the length of _rec_.
+          1. For each primitive data property _pdp_, do
```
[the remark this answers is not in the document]
rricard: meant to do it, forgot about it

spec/data-types-and-values.html
```diff
@@ -7,12 +7,35 @@ <h1>ECMAScript Language Types</h1>
...
+        <h1>RecordTypeCreate ( _keysList_, _valuesList_ )</h1>
+        <p>This abstract operation RecordTypeCreate takes a _keysList_ and a _valuesList_ argument. It is used to specify the creation of new Record value types. It performs the following steps when called:</p>
+        <emu-alg>
+            1. Assert: _keysList_ is a List of Strings.
+            1. Assert: _valuesList_ is a List of primitive values.
+
+            1. Assert: _valuesList_.length == _keysList_.length
+            1. Let _L_ be an empty List.
+            1. For each _i_ from 0 to _keysList_.length:
+              1. Add { [[Key]]: _keysList_[_i_], [[Value]]: _valuesList_[_i_] } to _L_
```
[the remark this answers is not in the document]
rricard: This is what this is trying to do, that is why it sorts afterwards

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -63,10 +63,9 @@ <h1>RecordCreate ( _value_ )</h1>
         <emu-alg>
             1. Assert: Type(_value_) is Record.
             1. Let _R_ be ! MakeBasicObject(&laquo; [[RecordData]] &raquo;).
+            1. Set _R_'s essential internal methods to the definitions specified <emu-xref href="#sec-record-exotic-objects"></emu-xref>.
+            1. Set _R_.[[Prototype]] to *null*
             1. Set _R_.[[RecordData]] to _value_
```
ljharb:
```suggestion
            1. Set _R_.[[Prototype]] to *null*.
            1. Set _R_.[[RecordData]] to _value_.
```

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,97 @@
...
+<emu-clause id="sec-ordinary-and-exotic-objects-behaviours">
+  <h1>Ordinary and Exotic Objects Behaviours</h1>
+
+  <emu-clause id="sec-built-in-exotic-object-internal-methods-and-slots">
+    <h1>Built-in Exotic Object Internal Methods and Slots</h1>
+    <emu-clause id="sec-record-exotic-objects">
+      <h1>Record Exotic Objects</h1>
+      <p>A Record object is an exotic object that encapsuates a Record value and exposes virtual string-indexed data properties corresponding to the individual keys set on the underlying Record value. All keys properties are non-writable and non-configurable.</p>
+
+      <p>An object is a <dfn id="record-exotic-object">Record exotic object</dfn> (or simply, a Record object) if its [[GetOwnPoperty]], [[DefineOwnProperty]], and [[OwnPropertyKeys]] internal methods.</p>
```
littledan, line 13: I guess this copypasta got cut off early.

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,97 @@
...
+      <emu-clause id="sec-recordcreate" aoid="RecordCreate">
+        <h1>RecordCreate ( _value_ )</h1>
+        <p>This abstract operation RecordCreate takes a _value_ argument. It is used to specify the creation of new Record exotic objects. It performs the following steps when called:</p>
+        <emu-alg>
+            1. Assert: Type(_value_) is Record.
+            1. Let _R_ be ! MakeBasicObject(&laquo; [[RecordData]] &raquo;).
+            1. Set _R_.[[RecordData]] to _value_
+            1. Set _R_.[[GetOwnProperty]] as specified in <emu-xref href="#sec-record-exotic-objects-getownproperty-p"></emu-xref>.
+            1. Set _R_.[[DefineOwnProperty]] as specified in <emu-xref href="#sec-record-exotic-objects-defineownproperty-p-desc"></emu-xref>.
+            1. Set _R_.[[OwnPropertyKeys]] as specified in <emu-xref href="#sec-record-exotic-objects-ownpropertykeys"></emu-xref>.
```
littledan: Some things missing here off-hand:
- You haven't done PreventExtensions and need to
- The prototype needs to be null somehow (either by declaring a [[Prototype]] slot and setting it to null, or by making SetPrototypeOf/GetPrototypeOf methods that act like it's null)

tbh I was picturing that all the methods would be overridden from scratch, like in [module namespace exotic objects](https://tc39.es/ecma262/#sec-module-namespace-exotic-objects), rather than falling back to ordinary object behavior like strings do. You're not using any of the object internal slots, though some things do "work out" with this phrasing anyway (such as [[HasProperty]]).

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,97 @@
...
+            1. Let _rec_ be _R_.[[RecordData]]
+            1. Assert Type(rec) is Record
+            1. Let _len_ be the length of _rec_.
+            1. Let _resultPrim_ be ~empty~
+            1. For each pair _p_ of _rec_.[[Fields]], do
+              1. Let _key_ be _p_.[[Key]]
+              1. Let _prim_ be _p_.[[Value]]
+              1. Assert: Type(_key_) is String
+              1. Assert: Type(_prim_) is not Object
+              1. If _key_ == _P_, let _resultPrim_ be _prim_
```
littledan, line 89: Avoid ==; spec text tends to use =, SameValue, is, or some wording like that. I think all three of those would mean the same thing here and each be acceptable.

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,97 @@
...
+
+      <emu-clause id="sec-record-exotic-objects-defineownproperty-p-desc">
+        <h1>[[DefineOwnProperty]] ( _P_, _Desc_ )</h1>
+        <p>When the [[DefineOwnProperty]] internal method of a Record exotic object _R_ is called with property key _P_, and Property Descriptor _Desc_ the following steps are taken:</p>
+        <emu-alg>
+          1. Assert: IsPropertyKey(_P_) is *true*.
+          1. Let _current_ be ? _R_.[[GetOwnProperty]](_P_).
+          1. If _current_ is *undefined* return *false*.
+          1. If IsAccessorDescriptor(_Desc_) is *true*, return *false*.
+          1. If _Desc_.[[Writable]] is present and has value *false*, return *false*.
```
littledan, line 36: I guess this condition is inverted

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,97 @@
...
+
+      <p>Record exotic objects have the same internal slots as ordinary objects. They also have a [[RecordData]] internal slot.</p>
+
+
+      <emu-clause id="sec-record-exotic-objects-getownproperty-p">
+        <h1>[[GetOwnProperty]] ( _P_ )</h1>
+        <p>When the [[GetOwnProperty]] internal method of a Record exotic object _R_ is called with property key _P_, the following steps are taken:</p>
+        <emu-alg>
+          1. Assert: IsPropertyKey(_P_) is *true*.
+          1. If Type(_P_) is a Symbol, throw a *TypeError* exception.
```
littledan, line 23: I think we should return *undefined* from this case.

spec/data-types-and-values.html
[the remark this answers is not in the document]
rricard: Ok, that has the consequence of defining another intermediary structure, are we ok with that?

spec/ordinary-and-exotic-object-behaviours.html
rricard: this is an outdated diff: I am in the process of getting things from module ns exotic object, still not done though

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -40,6 +55,43 @@ <h1>[[DefineOwnProperty]] ( _P_, _Desc_ )</h1>
...
+      <emu-clause id="sec-record-exotic-objects-set-p-receiver">
+        <h1>[[Set]] ( _P_, _Receiver_ )</h1>
+        <p>When the [[Set]] internal method of a Record exotic object _R_ is called with property key _P_ and ECMAScript language value _Receiver_, the following steps are taken:</p>
+        <emu-alg>
+          1. Return *false*.
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-record-exotic-objects-delete-p">
+        <h1>[[Delete]] ( _P_ )</h1>
```
littledan, line 91: Check the Delete method of module namespace exotic objects again

rricard:
I'm still working on it dan, will let you know when I need another review. A bunch of things are still in an incoherent state

spec/ordinary-and-exotic-object-behaviours.html
rricard: yes, it should return true for non-existing keys. I just did a first pass for now but I'm going to do a second one

spec/data-types-and-values.html
```diff
@@ -7,12 +7,63 @@ <h1>ECMAScript Language Types</h1>
...
+        <emu-alg>
+            1. Assert: _entries_ is a List.
+            1. Let _uniqueEntries_ be an empty Set.
+            1. For each _kv_ in _entries_:
+              1. Assert: _kv_.[[Key]] is a String.
+              1. Assert: _kv_.[[Value]] is a primitive type.
+              1. For each _existing_ in _uniqueEntries_:
+                1. If SameValue(_existing_.[[Key]], _kv_.[[Key]]
+                  1. Remove _existing_ from _uniqueEntries_
+                  1. Break
```
littledan, line 26: So far, we've been avoiding using "Break" in spec language, sticking to just early returns. How about instead, you iterate over _entries_ in reverse order, and do "If there is no entry _existing_ in _uniqueEntries_ such that _existing_.[[Key]] is _kv_.[[Key]], append _kv_ to _uniqueEntries_."

spec/data-types-and-values.html
```diff
@@ -7,12 +7,63 @@ <h1>ECMAScript Language Types</h1>
...
+              1. Add _kv_ to _uniqueEntries_
+            1. Let _sortedEntries_ be a new List containing the same values as the list _entries_ where the values are ordered as if an Array of the same values had been sorted using `Array.prototype.sort` using <emu-xref href="#sec-recordentrycompare">RecordEntryCompare</emu-xref> as _comparefn_.
```
littledan, line 28: Pedantically speaking, it seems inappropriate to use Array.prototype.sort on a List of Records--you can't put these Records in an Array. Also I think it would be clearer to use less indirection in this algorithm (though I can see how this was inspired by wording in module namespace exotic objects). How about this wording:

1. Sort _entries_ such that, for any entry two entries _a_ preceding _b_, then the result of performing Abstract Relational Comparison _a_.[[Key]] < _b_.[[Key]] is *true*.

spec/data-types-and-values.html
```diff
@@ -7,12 +7,63 @@ <h1>ECMAScript Language Types</h1>
...
+            1. Let _R_ be an empty Record
+            1. Set _R_.[[Field]] to _sortedEntries_
```
littledan, line 30: To avoid making it look like a primitive is being mutated (which wouldn't make sense), this could be reworded similarly to [the creation of Symbols](https://tc39.es/ecma262/#sec-symbol-description) as,

Let _R_ be a new Record value whose [[Fields]] value is _sortedEntries_

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,326 @@
...
+      
+      <emu-clause id="sec-recordhasproperty-r-p" aoid="RecordHasProperty">
+        <h1>RecordHasProperty ( _R_, _P_ )</h1>
+        <p>This abstract operation RecordHasProperty takes arguments _R_ and _P_. It performs the following steps when called:</p>
+        <emu-alg>
+          1. Assert: IsPropertyKey(_P_) is *true*.
+          1. If Type(_P_) is a Symbol, return *false*.
+          1. Let _rec_ be _R_.[[RecordData]].
+          1. Assert Type(_rec_) is a Record.
+          1. For each pair _p_ of _rec_.[[Fields]], do
```
littledan, line 137: Nit: s/pair/entry/

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,326 @@
...
+        </emu-alg>
+      </emu-clause>
+      
+      <emu-clause id="sec-recordhasproperty-r-p" aoid="RecordHasProperty">
+        <h1>RecordHasProperty ( _R_, _P_ )</h1>
+        <p>This abstract operation RecordHasProperty takes arguments _R_ and _P_. It performs the following steps when called:</p>
+        <emu-alg>
+          1. Assert: IsPropertyKey(_P_) is *true*.
+          1. If Type(_P_) is a Symbol, return *false*.
+          1. Let _rec_ be _R_.[[RecordData]].
```
littledan, line 135: Note, for Symbol [[Description]], there are some uses of "_sym_'s [[Description]]" and some of "_sym_.[[Description]]". We should decide on one or the other convention and use it across the Record and Tuple specs. So far, in #131 , the tendency has been the former, but I'd be open to switching this.

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,326 @@
...
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-recordcreate" aoid="RecordCreate">
+        <h1>RecordCreate ( _value_ )</h1>
+        <p>This abstract operation RecordCreate takes a _value_ argument. It is used to specify the creation of new Record exotic objects. It performs the following steps when called:</p>
+        <emu-alg>
+            1. Assert: Type(_value_) is a Record.
+            1. Assert: Type(_value_.[[Fields]]) is a List.
+            1. Let _R_ be ! MakeBasicObject(&laquo; [[RecordData]] &raquo;).
```
littledan, line 119: I guess you need [[Prototype]] here too

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,326 @@
...
+            1. Set _R_'s essential internal methods to the definitions specified in <emu-xref href="#sec-record-exotic-objects"></emu-xref>.
+            1. Set _R_.[[Prototype]] to *null*.
+            1. Set _R_.[[RecordData]] to _value_.
+            1. OrdinaryPreventExtensions(_R_).
```
littledan, line 123: This line should be deleted; it doesn't make sense (since Record wrappers are not ordinary objects) and is unnecessary (since it already implements the extension-related methods appropriately).

spec/ordinary-and-exotic-object-behaviours.html
littledan: Note, this is still a TODO

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,326 @@
...
+          1. Let _rec_ be _R_.[[RecordData]].
+          1. Assert Type(_rec_) is a Record.
+          1. Let _len_ be the length of _rec_.
+          1. Let _resValue_ be *undefined*.
+          1. For each pair _p_ of _rec_.[[Fields]], do:
+            1. Let _key_ be _p_.[[Key]].
+            1. Let _val_ be _p_.[[Value]].
+            1. Assert: Type(_key_) is a String.
+            1. Assert: Type(_val_) is not an Object.
+            1. If SameValue(_key_, _P_), let _resValue_ be _val_.
```
littledan, line 159: Note, it's fine to early-return from an algorithm, so you could "return _val_" here, and fall back to returning undefined below.

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,326 @@
...
+          1. Return _resValue_.
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-recordgetownproperty-r-p" aoid="RecordGetOwnProperty">
+        <h1>RecordGetOwnProperty ( _R_, _P_ )</h1>
```
littledan, line 165: Nit: I don't see much of a reason to spit out RecordGetOwnProperty into a separate algorithm, rather than leaving it simply in the method definitions above (which is how the rest of the spec tends to do it). Even for RecordGet and RecordHas, which seem like they will need to be called from other algorithms, you could just invoke them as their methods on the object from other algorithms (other types of objects seem to do this).

spec/data-types-and-values.html
```diff
@@ -7,12 +7,63 @@ <h1>ECMAScript Language Types</h1>
...
+      <p>
+        The Record type is the set of all finite mappings from Strings to ECMAScript primitive values including Record and Tuple. Each record value holds an associated [[Fields]] List value which is a list of pairs of the form { [[Key]], [[Value]] } where the [[Key]] is a String and [[Value]] is any primitive value. Entries of [[Fields]] are sorted by [[Key]] in code unit order. The [[Fields]] List and its entries are never modified.
+     </p>
+
+      <emu-clause id="sec-makerecordvalue" aoid="MakeRecordValue">
+        <h1>MakeRecordValue ( _entries_ )</h1>
+        <p>This abstract operation MakeRecordValue takes a _entries_ argument. It is used to specify the creation of new Record value types. It performs the following steps when called:</p>
+        <emu-alg>
+            1. Assert: _entries_ is a List.
+            1. Let _uniqueEntries_ be an empty Set.
```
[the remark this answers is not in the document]
rricard: Meant to be transformed back to list

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,326 @@
...
+            1. Add _key_ to _keys_
+          1. Return _keys_
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-recordcreate" aoid="RecordCreate">
+        <h1>RecordCreate ( _value_ )</h1>
+        <p>This abstract operation RecordCreate takes a _value_ argument. It is used to specify the creation of new Record exotic objects. It performs the following steps when called:</p>
+        <emu-alg>
+            1. Assert: Type(_value_) is a Record.
```
ljharb, line 117: why would RecordCreate take a Record? This seems like more the domain of https://tc39.es/ecma262/#sec-toobject than of its own abstract op

littledan: I was a little mystified by the naming convention at first, but see analogous operations like https://tc39.es/ecma262/#sec-stringcreate and https://tc39.es/ecma262/#sec-modulenamespacecreate

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,326 @@
...
+
+      <p>Record exotic objects have the same internal slots as ordinary objects. They also have a [[RecordData]] internal slot.</p>
+
+      <emu-clause id="sec-record-exotic-objects-setprototypeof-v">
+        <h1>[[SetPrototypeOf]] ( _V_ )</h1>
+        <p>When the [[SetPrototypeOf]] internal method of a Record exotic object _R_ is called with argument _V_, the following steps are taken:</p>
+        <emu-alg>
+          1. Return ! SetImmutablePrototype(_R_, _V_).
+        </emu-alg>
+      </emu-clause>
```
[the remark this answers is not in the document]
littledan: I was thinking this should follow what other exotic objects do, of defining this as a method, see https://tc39.es/ecma262/#sec-module-namespace-exotic-objects-setprototypeof-v . I'm not really sure which wording to use from what you're quoting, since it doesn't have the internal methods defined for ordinary objects.

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,326 @@
...
+      
+      <emu-clause id="sec-record-exotic-objects-isextensible">
+        <h1>[[IsExtensible]] ()</h1>
+        <p>When the [[IsExtensible]] internal method of a Record exotic object _R_ is called, the following steps are taken:</p>
+        <emu-alg>
+          1. Return *false*.
+        </emu-alg>
+      </emu-clause>
```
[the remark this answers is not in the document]
littledan: I'm not sure what you mean; the idea is that this wrapper doesn't have any of the object internal slots, including [[Extensible]]. See https://tc39.es/ecma262/#sec-module-namespace-exotic-objects-isextensible for analogous spec text.

spec/ordinary-and-exotic-object-behaviours.html
ljharb: hmm, strange that String has it but Number, Boolean, etc don't (objects having a FooCreate makes sense to me; but primitives don't). It feels like the precedent is more for primitives not to have one.

littledan: I guess what distinguishes String is that its wrapper is exotic, like Record. Maybe we should continue this discussion in an editorial issue in ecma262? I want to make sure we have clear text somewhere or other to explain how the wrapper is set up, and I think it should match String, but I am fine with moving to some other way of organizing the text.

spec/data-types-and-values.html
rricard: We don't necessarily need to break, but I will try your approach

spec/data-types-and-values.html
rricard: Got my inspiration from here: https://tc39.es/ecma262/#sec-modulenamespacecreate

I understand it looks heavy but this seems to do what we want. I'll explore Abstract Relational Comparison later once I'm done with the rest.

spec/ordinary-and-exotic-object-behaviours.html
rricard: I prefer the dot notation as it is less noisy `.` vs. `'s ` but I'm open to change as well

spec/ordinary-and-exotic-object-behaviours.html
rricard: I put those functions here as a temporary location, they are not meant to stay there

spec/data-types-and-values.html
```diff
@@ -7,12 +7,59 @@ <h1>ECMAScript Language Types</h1>
...
+
+    <emu-clause id="sec-maketuplevalue" aoid="MakeTupleValue">
+      <h1>MakeTupleValue ( _values_ )</h1>
+      <p>This abstract operation MakeTupleValue takes a _values_ argument. It is used to specify the creation of new Tuple value types. It performs the following steps when called:</p>
+      <emu-alg>
+          1. Assert: _values_ is a List.
+          1. For each _value_ in _values_:
+            1. Assert: Type(_value_) is not Object.
+          1. Let _T_ be an empty Tuple
+          1. Set _T_.[[Sequence]] to _values_
```
littledan, line 59: Use wording here analogous to line 26.

spec/data-types-and-values.html
```diff
@@ -7,12 +7,59 @@ <h1>ECMAScript Language Types</h1>
...
+        <p>This abstract operation MakeRecordValue takes a _entries_ argument. It is used to specify the creation of new Record value types. It performs the following steps when called:</p>
+        <emu-alg>
+            1. Assert: _entries_ is a List.
+            1. Let _reversedEntries_ be a new List containing the same values as the list _entries_ where the values are ordered as an Array of the same values has been sorted using `Array.prototype.reverse`.
+            1. Let _uniqueEntries_ be an empty List.
+            1. For each _kv_ in _reversedEntries_:
+              1. Assert: Type(_kv_.[[Key]]) is String.
+              1. Assert: Type(_kv_.[[Value]]) is not Object.
+              1. If there is no entry _existing_ in _uniqueEntries_ such that _existing_.[[Key]] is _kv_.[[Key]], append _kv_ to _uniqueEntries_.
+            1. Let _sortedEntries_ be a new List containing the same values as the list _uniqueEntries_ where the values are ordered as if an Array of the same values had been sorted using `Array.prototype.sort` using <emu-xref href="#sec-recordentrycompare">RecordEntryCompare</emu-xref> as _comparefn_.
```
littledan, line 25: This language still needs to change, as I described in https://github.com/tc39/proposal-record-tuple/pull/132#discussion_r445150000 . I understand that you tried to copy it from that place, but it doesn't make sense to apply that text here, as it is sorting strings (which are JS values) and you're sorting records here (which aren't) and passing in algorithm in place of a function (which is also invalid). I suggest you use the wording I provided to start, and we can iterate on it later.

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,316 @@
...
+
+      <emu-clause id="sec-record-exotic-objects-getownproperty-p">
+        <h1>[[GetOwnProperty]] ( _P_ )</h1>
+        <p>When the [[GetOwnProperty]] internal method of a Record exotic object _R_ is called with property key _P_, the following steps are taken:</p>
+        <emu-alg>
+          1. Assert: IsPropertyKey(_P_) is *true*.
+          1. If Type(_P_) is a Symbol, return *undefined*.
+          1. Let _getResult_ be ! RecordGet(_R_, _P_).
+          1. If _getResult_ is *undefined*, return *undefined*.
```
littledan, line 40: It's important to distinguish undefined from missing here, as the value in the record may actually be undefined. You can use `~empty~` as a distinguished return value of RecordGet for this case, for example.

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,316 @@
...
+          1. Assert: Type(_P_) is String.
+          1. Let _rec_ be _R_.[[RecordData]].
+          1. Assert Type(_rec_) is Record.
+          1. Let _len_ be the length of _rec_.
+          1. For each _entry_ of _rec_.[[Fields]], do:
+            1. Let _key_ be _entry_.[[Key]].
+            1. Let _val_ be _entry_.[[Value]].
+            1. Assert: Type(_key_) is String.
+            1. Assert: Type(_val_) is not Object.
+            1. Return _val_.
```
littledan, line 162: I guess you're missing a comparison between _key_ and _P_ here, and then a return of `~empty~` outside the loop if you didn't reach an early return.

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,316 @@
...
+
+      <emu-clause id="sec-tuple-exotic-objects-getownproperty-p">
+        <h1>[[GetOwnProperty]] ( _P_ )</h1>
+        <p>When the [[GetOwnProperty]] internal method of a Tuple exotic object _T_ is called with property key _P_, the following steps are taken:</p>
+        <emu-alg>
+          1. Assert: IsPropertyKey(_P_) is *true*.
+          1. If Type(_P_) is String, then
+            1. Let _numericIndex_ be ! CanonicalNumericIndexString(_P_).
+            1. If _numericIndex_ is not *undefined*, then
+              1. Let _value_ be ? TupleGet(_T_, _numericIndex_)
```
littledan, line 200: Again, you should make sure that TupleGet distinguishes undefined from missing values, as Tuples may contain undefined.

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,316 @@
...
+          1. Assert: IsPropertyKey(_P_) is *true*.
+          1. If Type(_P_) is String, then
+            1. Let _numericIndex_ be ! CanonicalNumericIndexString(_P_).
+            1. If _numericIndex_ is not *undefined*, then
+              1. Let _current_ be ? TupleGet(_T_, _numericIndex_)
+              1. If _current_ is *undefined* return *false*.
+              1. If IsAccessorDescriptor(_Desc_) is *true*, return *false*.
+              1. If _Desc_.[[Writable]] is present and has value *true*, return *false*.
+              1. If _Desc_.[[Enumerable]] is present and has value *false*, return *false*.
+              1. If _Desc_.[[Configurable]] is present and has value *true*, return *false*.
```
littledan, line 233: Editorial suggestion: These four conditionals are to be checked regardless of what TupleGet outputs, so I think it makes sense to put them before the TupleGet call.

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,316 @@
...
+            
+      <emu-clause id="sec-tuple-exotic-objects-get-p-receiver">
+        <h1>[[Get]] ( _P_, _Receiver_ )</h1>
+        <p>When the [[Get]] internal method of a Tuple exotic object _T_ is called with property key _P_ and ECMAScript language value _Receiver_, the following steps are taken:</p>
+        <emu-alg>
+          1. If Type(_P_) is String, then
+            1. Let _numericIndex_ be ! CanonicalNumericIndexString(_P_).
+            1. If _numericIndex_ is not *undefined*, then
+              1. Return ? TupleGet(_T_, _numericIndex_)
+          1. Return *undefined*
```
littledan, line 248: This should fall back to the prototype at this point.

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,316 @@
...
+
+      
+      <emu-clause id="sec-tuplecreate" aoid="TupleCreate">
+        <h1>TupleCreate ( _value_ )</h1>
+        <p>This abstract operation TupleCreate takes a _value_ argument. It is used to specify the creation of new Tuple exotic objects. It performs the following steps when called:</p>
+        <emu-alg>
+            1. Assert: Type(_value_) is Tuple.
+            1. Assert: Type(_value_.[[Sequence]]) is List.
+            1. Let _T_ be ! MakeBasicObject(&laquo; [[TupleData]], [[Prototype]] &raquo;).
+            1. Let _T_.[[TupleLength]] be _value_.[[Sequence]].length
```
littledan, line 281: I don't think there's any reason to store the TupleLength like this. Better to just figure it out from _value_.[[Sequence]] when you need it. You could factor that out into an algorithm if you need it--that would be preferable to making a "denormalized" data structure like this. (Also, the length of Lists is not denoted by `.length` but rather "the length of _value_.[[Sequence]]")

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,316 @@
...
+            1. Set _T_'s essential internal methods to the definitions specified in <emu-xref href="#sec-tuple-exotic-objects"></emu-xref>.
+            1. Set _T_.[[Prototype]] to *null*.
```
littledan, line 283: The prototype is %Tuple.prototype%, not null

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,316 @@
...
+      </emu-clause>
+      
+      <emu-clause id="sec-tupleget-r-p" aoid="TupleGet">
+        <h1>TupleGet ( _T_, _numericIndex_ )</h1>
+        <p>This abstract operation TupleGet takes arguments _T_ and _numericIndex_. It performs the following steps when called:</p>
+        <emu-alg>
+          1. Assert: _T_ is an Object that has [[TupleData]] internal slot.
+          1. Assert: Type(_numericIndex_) is an integer.
+          1. Let _tup_ be _T_.[[TupleData]].
+          1. If ! IsValidTupleIndex(_numericIndex_) is *false*, return *undefined*.
```
littledan, line 309: In this case, return `~empty~` instead. (Or, assert it's true and expect callers to call IsValidTupleIndex.)

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,316 @@
...
+      <emu-clause id="sec-tuple-exotic-objects-hasproperty-p">
+        <h1>[[HasProperty]] ( _P_ )</h1>
+        <p>When the [[HasProperty]] internal method of a Tuple exotic object _T_ is called with property key _P_, the following steps are taken:</p>
+        <emu-alg>
+          1. Assert: IsPropertyKey(_P_) is *true*.
+          1. If Type(_P_) is String, then
+            1. Let _numericIndex_ be ! CanonicalNumericIndexString(_P_).
+            1. If _numericIndex_ is not *undefined*, then
+              1. Return ? IsValidTupleIndex(_T_, _numericIndex_)
+          1. Return *false*
```
littledan, line 216: We need to fall back to the prototype here.

spec/data-types-and-values.html
rricard: Got it!

spec/ordinary-and-exotic-object-behaviours.html
rricard: Completely forgot that, thanks!

rricard:
Thanks Dan, I will try to get around it soon!

spec/ordinary-and-exotic-object-behaviours.html
rricard: Synced up with rick, he's using dot notation

spec/data-types-and-values.html
```diff
@@ -7,12 +7,58 @@ <h1>ECMAScript Language Types</h1>
...
+      <p>
+        The Record type is the set of all finite mappings from Strings to ECMAScript primitive values including Record and Tuple. Each record value holds an associated [[Fields]] List value which is a list of pairs of the form { [[Key]], [[Value]] } where the [[Key]] is a String and [[Value]] is any primitive value. Entries of [[Fields]] are sorted by [[Key]] in code unit order. The [[Fields]] List and its entries are never modified.
+     </p>
+
+      <emu-clause id="sec-makerecordvalue" aoid="MakeRecordValue">
+        <h1>MakeRecordValue ( _entries_ )</h1>
+        <p>This abstract operation MakeRecordValue takes a _entries_ argument. It is used to specify the creation of new Record value types. It performs the following steps when called:</p>
+        <emu-alg>
+            1. Assert: _entries_ is a List.
+            1. Let _reversedEntries_ be a new List containing the same values as the list _entries_ where the values are ordered as an Array of the same values has been sorted using `Array.prototype.reverse`.
```
ljharb, line 19: this "as if" prose is a bit ambiguous with the mention of Array.prototype.reverse, which both doesn't sort and also causes observable lookups; i'd prefer either prose that describes reversing the list, or, just iterating it in reverse?

spec/data-types-and-values.html
ljharb: similarly here; delegating to a public method doesn't seem like an ideal approach to me.

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,324 @@
...
+      </emu-clause>
+
+      <emu-clause id="sec-record-exotic-objects-getownproperty-p">
+        <h1>[[GetOwnProperty]] ( _P_ )</h1>
+        <p>When the [[GetOwnProperty]] internal method of a Record exotic object _R_ is called with property key _P_, the following steps are taken:</p>
+        <emu-alg>
+          1. Assert: IsPropertyKey(_P_) is *true*.
+          1. If Type(_P_) is a Symbol, return *undefined*.
+          1. Let _getResult_ be ! RecordGet(_R_, _P_).
+          1. If _getResult_ is ~empty~, return *undefined*.
```
ljharb, line 40: does this mean i can't set an own expando string property on a boxed Record? `var o = Object('a'); o.foo = true; o.foo;` works, for example.

spec/data-types-and-values.html
rricard: Yes I just simplified the prose!

spec/data-types-and-values.html
```diff
@@ -16,28 +16,18 @@ <h1>MakeRecordValue ( _entries_ )</h1>
         <p>This abstract operation MakeRecordValue takes a _entries_ argument. It is used to specify the creation of new Record value types. It performs the following steps when called:</p>
         <emu-alg>
             1. Assert: _entries_ is a List.
-            1. Let _reversedEntries_ be a new List containing the same values as the list _entries_ where the values are ordered as an Array of the same values has been sorted using `Array.prototype.reverse`.
+            1. Reverse _entries_.
```
ljharb, line 19: this seems to be mutating the list; it'd be great to avoid that.

spec/ordinary-and-exotic-object-behaviours.html
rricard: That is the intent here, if that is an issue for you we should probably discuss that in an issue. If that is a blocker I'd be happy to enable additional properties in the spec and wait for Dan's return to discuss it more in depth

spec/data-types-and-values.html
rricard: What about something like:

1. Let _sortedEntries_ be _entries_ in reverse.

rricard: I'll do the same for _sortedEntries_

spec/ordinary-and-exotic-object-behaviours.html
rricard: Creating an issue, this is worth discussing either way. I think I'll do a first draft as is and we can update that once that issue progresses.

rricard: See # 137

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,322 @@
...
+      </emu-clause>
+      
+      <emu-clause id="sec-tuple-exotic-objects-ownpropertykeys">
+        <h1>[[OwnPropertyKeys]] ( )</h1>
+        <p>When the [[OwnPropertyKeys]] internal method of a Record exotic object _R_ is called, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _keys_ be a new empty List.
+          1. Assert _T_ is Tuple exotic object.
+          1. Let _len_ be the length of _T_.[[TupleData]].[[Sequence]].
+          1. For each integer _i_ starting with 0 such that _i_ &lt; _len_ in ascending order, do
```
ljharb, line 274: this reads awkwardly to me; usually this is done with an i variable, set to 0, and a while loop.

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,322 @@
...
+        <p>This abstract operation TupleCreate takes a _value_ argument. It is used to specify the creation of new Tuple exotic objects. It performs the following steps when called:</p>
+        <emu-alg>
+            1. Assert: Type(_value_) is Tuple.
+            1. Assert: Type(_value_.[[Sequence]]) is List.
+            1. Let _T_ be ! MakeBasicObject(&laquo; [[TupleData]], [[Prototype]] &raquo;).
+            1. Set _T_'s essential internal methods to the definitions specified in <emu-xref href="#sec-tuple-exotic-objects"></emu-xref>.
+            1. Set _T_.[[Prototype]] to %Tuple.prototype%.
+            1. Set _T_.[[TupleData]] to _value_.
+            1. Return _T_
+
```
ljharb, line 292:
```suggestion
            1. Return _T_.
```

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,322 @@
...
+            
+        </emu-alg>
+      </emu-clause>
+      
+      <emu-clause id="sec-isvalidtupleindex-r-p" aoid="IsValidTupleIndex">
+        <h1>IsValidTupleIndex ( _T_, _numericIndex_ )</h1>
+        <p>This abstract operation IsValidTupleIndex takes arguments _T_ and _numericIndex_. It performs the following steps when called:</p>
+        <emu-alg>
+          1. Assert: _T_ is an Object that has [[TupleData]] internal slot.
+          1. Assert: Type(_numericIndex_) is an integer.
```
ljharb, line 301: "integer" isn't a type. you may want this:
```suggestion
          1. Assert: ! IsInteger(_numericIndex_) is *true*.
```

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,322 @@
...
+          1. If _numericIndex_ is *-0*, return *false*
```
ljharb, line 302:
```suggestion
          1. If _numericIndex_ is *-0*, return *false*.
```

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,322 @@
...
+      <emu-clause id="sec-tupleget-r-p" aoid="TupleGet">
+        <h1>TupleGet ( _T_, _numericIndex_ )</h1>
+        <p>This abstract operation TupleGet takes arguments _T_ and _numericIndex_. It performs the following steps when called:</p>
+        <emu-alg>
+          1. Assert: _T_ is an Object that has [[TupleData]] internal slot.
+          1. Assert: Type(_numericIndex_) is an integer.
+          1. Let _tup_ be _T_.[[TupleData]].
+          1. If ! IsValidTupleIndex(_numericIndex_) is *false*, return ~empty~.
+          1. Let _value_ be the element with index _numericIndex_ in _tup_.[[Sequence]]
+          1. Return _value_
```
ljharb, line 317:
```suggestion
          1. Let _value_ be the element with index _numericIndex_ in _tup_.[[Sequence]].
          1. Return _value_.
```

spec/ordinary-and-exotic-object-behaviours.html
rricard: I can change that but FYI, you might consider an editorial change here: https://tc39.es/ecma262/#sec-string-exotic-objects-ownpropertykeys as I just took the same formulation

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,322 @@
...
+          1. If Type(_P_) is a Symbol, return *undefined*.
+          1. Let _value_ be ! RecordGet( _R_, _P_ ).
+          1. If _value_ is ~empty~, return *undefined*.
+          1. Return _value_
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-record-exotic-objects-set-p-receiver">
+        <h1>[[Set]] ( _P_, _Receiver_ )</h1>
+        <p>When the [[Set]] internal method of a Record exotic object _R_ is called with Property Key _P_ and ECMAScript language value _Receiver_, the following steps are taken:</p>
```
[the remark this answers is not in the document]
rricard: I think I'm gonna allow symbols to be set in that version of the spec but still not support string props just yet until we solve #137

spec/data-types-and-values.html
```diff
@@ -7,12 +7,48 @@ <h1>ECMAScript Language Types</h1>
...
+        <p>This abstract operation MakeRecordValue takes a _entries_ argument. It is used to specify the creation of new Record value types. It performs the following steps when called:</p>
+        <emu-alg>
+            1. Assert: _entries_ is a List.
+            1. Let _reversedEntries_ be _entries_ in reverse.
+            1. Let _uniqueEntries_ be an empty List.
+            1. For each _kv_ in _reversedEntries_:
+              1. Assert: Type(_kv_.[[Key]]) is String.
+              1. Assert: Type(_kv_.[[Value]]) is not Object.
+              1. If there is no entry _existing_ in _uniqueEntries_ such that _existing_.[[Key]] is _kv_.[[Key]], append _kv_ to _uniqueEntries_.
+            1. Let _sortedEntries_ be _uniqueEntries_ sorted such that, for any entry two entries _a_ preceding _b_, then the result of performing Abstract Relational Comparison _a_.[[Key]] < _b_.[[Key]] is *true*.
```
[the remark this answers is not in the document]
rricard: I think using `<` would be more accurate than localeCompare as localeCompare would have different results in different locales, what I am trying to express is char code ordering, I will attempt another change to the wording but if that is not good enough to you, is it possible 
to suggest something that would express that?

spec/ordinary-and-exotic-object-behaviours.html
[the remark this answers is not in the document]
littledan: We discussed this in committee and I explained why I think we should continue to prohibit defining symbols on Record wrappers. There is no way we could let the definition of these symbols be reflected on the underlying record--it would be superficially on the wrapper. I don't think we should encourage these wrappers to be used in a stateful way, and I don't think it would be very useful (e.g., equality doesn't work). You can always put a Proxy around a record wrapper if you want to add other behaviors like this.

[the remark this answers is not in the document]
rricard: Ok, I'm reverting that change, will discuss further in #137. Either way is easy to change in the spec so I don't want to block this PR because of it.

spec/data-types-and-values.html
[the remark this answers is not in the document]
rickbutton: How about:

1. Let _sortedEntries_ be _uniqueEntries_ sorted such that, for any two integers _i_, _j_ less than the length of list _uniqueEntries_, If _i_ < _j_ then the result of performing Abstract Relational Comparison _uniqueEntries_[_i_].[[Key]] < _uniqueEntries_[_j_].[[Key]] is *true*.

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,325 @@
...
+          1.Return ~empty~.
+        </emu-alg>
+      </emu-clause>
+    </emu-clause>
+
+    <emu-clause id="sec-tuple-exotic-objects">
+      <h1>Tuple Exotic Objects</h1>
+      <p>A Tuple object is an exotic object that encapsuates a Tuple value and exposes virtual integer-indexed data properties corresponding to the individual entries set on the underlying Tuple value. All keys properties are non-writable and non-configurable.</p>
+
+      <p>An object is a <dfn id="tuple-exotic-object">Tuple exotic object</dfn> (or simply, a Record object) if its following internal methods use the following implementations and is an <emu-xref href="#sec-immutable-prototype-exotic-objects">Immutable Prototype Exotic Object</emu-xref>..</p>
```
rickbutton, line 169:
```suggestion
      <p>An object is a <dfn id="tuple-exotic-object">Tuple exotic object</dfn> (or simply, a Tuple object) if its following internal methods use the following implementations and is an <emu-xref href="#sec-immutable-prototype-exotic-objects">Immutable Prototype Exotic Object</emu-xref>..</p>
```

spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -0,0 +1,325 @@
...
+        <emu-alg>
+          1. Let _keys_ be a new empty List.
+          1. Let _tup_ be _T_.[[TupleData]].
+          1. Assert: Type(_tup_) is Tuple.
+          1. Let _len_ be the length of _tup_.[[Sequence]].
+          1. Let _index_ be 0.
+          1. Repeat, while _index_ &lt; _len_,
+            1. Add ! ToString(_index_) to _keys_.
+            1. Set _index_ to _index_ + 1.
+          1. Let _ordinaryProps_ be ? OrdinaryOwnPropertyKeys (_R_).
```
rickbutton, line 277:
```suggestion
        <p>When the [[OwnPropertyKeys]] internal method of a Record exotic object _T_ is called, the following steps are taken:</p>
        <emu-alg>
          1. Let _keys_ be a new empty List.
          1. Let _tup_ be _T_.[[TupleData]].
          1. Assert: Type(_tup_) is Tuple.
          1. Let _len_ be the length of _tup_.[[Sequence]].
          1. Let _index_ be 0.
          1. Repeat, while _index_ &lt; _len_,
            1. Add ! ToString(_index_) to _keys_.
            1. Set _index_ to _index_ + 1.
          1. Let _ordinaryProps_ be ? OrdinaryOwnPropertyKeys (_T_).
```

rricard:
Need to solve tha merge conflict

spec/ordinary-and-exotic-object-behaviours.html
rickbutton: Please open an issue if you think this should still be changed @ljharb , we can discuss the choices in more detail.

PR #135, opened 2020-06-26
Record and Tuple abstract ops

rickbutton:
Draft PR. Comments welcome.

### DIFF
--- a/spec/abstract-operations.html
+++ b/spec/abstract-operations.html
@@ -388,7 +388,7 @@
               <ins>Record</ins>
             </td>
             <td>
-              <ins>Return ! RecordToString(_argument_).</ins>
+              <ins>Return ! Record::toString(_argument_).</ins>
             </td>
           </tr>
           <tr>
@@ -396,7 +396,7 @@
               <ins>Tuple</ins>
             </td>
             <td>
-              <ins>Return ! TupleToString(_argument_).</ins>
+              <ins>Return ! Tuple::toString(_argument_).</ins>
             </td>
           </tr>
           <tr>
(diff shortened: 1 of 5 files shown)


spec/data-types-and-values.html
```diff
@@ -8,11 +8,50 @@ <h1>ECMAScript Language Types</h1>
     <emu-clause id="sec-ecmascript-language-types-record-type">
       <h1>The Record Type</h1>
       <p>TBD</p>
+
+      <emu-clause id="sec-recordtostring" aoid="RecordToString">
+        <h1>RecordToString ( _argument_ )</h1>
+        <p>The abstract operation RecordToString takes argument _argument_ (a Record) and convertes _argument_ to a String. It performs the following steps when called:</p>
+        <emu-alg>
+            1. Assert: Type(_argument_) is Record.
+            1. Return *"[record]"*.
```
ljharb, line 17: that's it? no string version of the record, for debugging?

spec/immutable-data-structures.html
```diff
@@ -568,6 +568,40 @@ <h1>Tuple.prototype.unshifted ( ..._args_ )</h1>
...
+        <p>When the *toLocaleString* method is called, The following steps are taken:</p>
+        <emu-alg>
+          1. Let _T_ be ? thisTupleValue(*this* value).
+          1. Let _separator_ be the String value for the list-separator String appropriate for the host environment's current locale (this is derived in an implementation-defined way).
+          1. Let _steps_ be the algorithm steps defined in <emu-xref href="#sec-tuple-element-tolocalestring-functions" title></emu-xref>.
+          1. Let _mapper_ be ! CreateBuiltinFunction(_steps_, &laquo; &raquo;).
+          1. Return ? TupleToString(_T_, _separator_, _mapper_).
+        </emu-alg>
+
+        <emu-clause id="sec-tuple-element-tolocalestring-functions">
```
rickbutton, line 584: very much winging it here, would love some more input. Because records don't have a prototype, we won't have a `toLocaleString` to call on a Record object inside a Tuple. I'm special casing this here to use a mapping function that handles Record-type elements specifically, so that Tuples with Records won't fail on `toLocaleString`.

Is this the right way to do this?

spec/data-types-and-values.html
```diff
@@ -8,11 +8,50 @@ <h1>ECMAScript Language Types</h1>
...
       <h1>The Tuple Type</h1>
       <p>TBD</p>
+
+      <emu-clause id="sec-tupletostring" aoid="TupleToString">
+        <h1>TupleToString ( _argument_, [ _separator_, [ _mapper_ ] ] )</h1>
+        <p>The abstract operation TupleToString takes arguments _argument_, _separator_, and _mapper_ (a function object). _mapper_ will be invoked with *undefined* as the receiver. It performs the following steps when called:</p>
+        <emu-alg>
+            1. Assert: Type(_argument_) is Tuple.
+            1. Assert: _mapper_ is *undefined* or IsCallable(_mapper_) is *true*.
+            1. Assert: Type(_separator_) is Undefined or String.
```
ljharb, line 32:
```suggestion
        <p>The abstract operation TupleToString takes arguments _argument_ (a Tuple), _separator_ (a String), and _mapper_ (a function object). _mapper_ will be invoked with *undefined* as the receiver. It performs the following steps when called:</p>
        <emu-alg>
            1. Assert: _mapper_ is *undefined* or IsCallable(_mapper_) is *true*.
```
i'm not sure the IsCallable assertion is needed either, given that the prose says "function object"

spec/data-types-and-values.html
```diff
@@ -8,11 +8,50 @@ <h1>ECMAScript Language Types</h1>
...
+            1. Let _R_ be the empty String.
+            1. If _separator_ is *undefined*,
+              1. Let _realSep_ be *","*.
+            1. Else,
+              1. Let _realSep_ be _separator_.
+            1. Repeat, while _k_ &lt; _len_,
+              1. If _k_ &gt; 0, then
+                1. Set _R_ to the string-concatenation of _R_ and _realSep_.
+              1. Let _item_ be _list_[_k_].
+              If _mapper_
```
ljharb, line 45: a lot of this seems like it could/should be an abstract operation, shared by both this and https://tc39.es/ecma262/#sec-array.prototype.join ?

spec/abstract-operations.html
```diff
@@ -0,0 +1,673 @@
...
+          1. <ins>Let _yList_ be _y_'s [[Sequence]] value.</ins>
+          1. <ins>Let _xLen_ be the length of List _xList_.</ins>
+          1. <ins>Let _yLen_ be the length of List _yList_.</ins>
+          1. <ins>If _xLen_ &ne; _yLen_, Return *false*.</ins>
+          1. <ins>Let _k_ be 0.</ins>
+          1. <ins>Repeat, while _k_ &lt; _xLen_,</ins>
+            1. <ins>Let _xItem_ be _xList_[_k_].</ins>
+            1. <ins>Let _yItem_ be _yList_[_k_].</ins>
+            1. <ins>Let _same_ be ! SameValueZero(_xItem_, _yItem_).</ins>
+            1. <ins>If _same_ is *false*, Return *false*.</ins>
```
littledan, line 646: Nit: you could consolidate these two onto the same line, and use complex expressions if you want.

spec/abstract-operations.html
```diff
@@ -0,0 +1,673 @@
...
+        1. If Type(_x_) is Undefined, return *true*.
+        1. If Type(_x_) is Null, return *true*.
+        1. If Type(_x_) is String, then
+          1. If _x_ and _y_ are exactly the same sequence of code units (same length and same code units at corresponding indices), return *true*; otherwise, return *false*.
+        1. If Type(_x_) is Boolean, then
+          1. If _x_ and _y_ are both *true* or both *false*, return *true*; otherwise, return *false*.
+        1. If Type(_x_) is Symbol, then
+          1. If _x_ and _y_ are both the same Symbol value, return *true*; otherwise, return *false*.
+        1. <ins>If Type(_x_) is Tuple, then</ins>
+          1. <ins>Let _xList_ be _x_'s [[Sequence]] value.</ins>
```
littledan, line 636: Maybe we want to standardize on the _x_.[[Sequence]] notation used in #132 , in place of _x_'s [[Sequence]] value, but let's just choose one and use it consistently.

spec/abstract-operations.html
```diff
@@ -0,0 +1,673 @@
...
+          1. <ins>Let _yList_ be _y_'s [[Sequence]] value.</ins>
+          1. <ins>Let _xLen_ be the length of List _xList_.</ins>
+          1. <ins>Let _yLen_ be the length of List _yList_.</ins>
+          1. <ins>If _xLen_ &ne; _yLen_, Return *false*.</ins>
+          1. <ins>Let _k_ be 0.</ins>
+          1. <ins>Repeat, while _k_ &lt; _xLen_,</ins>
+            1. <ins>Let _xItem_ be _xList_[_k_].</ins>
+            1. <ins>Let _yItem_ be _yList_[_k_].</ins>
+            1. <ins>Let _same_ be ! SameValueZero(_xItem_, _yItem_).</ins>
```
littledan, line 645: I think this algorithm is part of the SameValue definition, so you probably want to use SameValue, not SameValueZero, here.

spec/immutable-data-structures.html
littledan: I don't see why you should handle Tuples with Records in them specially; I think it's fine if this fails in that case. We're just trying to be parallel to Arrays, not trying to make everything magical. I think the most important case here is really lists of numbers.

spec/data-types-and-values.html
ljharb: It definitely wouldn't be fine to look up the "join" property on Array.prototype :-) i'm suggesting it should have the same semantics via a shared abstract operation.

spec/data-types-and-values.html
ljharb: Objects return `[object Object]`, if you want it to be `[object Record]`, that's at least defensible.

For arrays, toString returns the contents, which is better than nothing - but either way this seems like an opportunity to do *better* than objects and arrays.

spec/data-types-and-values.html
littledan: Sure, but I think calling the original value %Array.prototype.join% would be safe as well and do what we want here.

spec/abstract-operations.html
ljharb: Yes please; internal slots are referenced with dots.

spec/data-types-and-values.html
littledan: Wanna open up an issue to track discussion? I'd be fine with `[object Record]` here as a starting point, personally.

ljharb: Filed #136.

spec/data-types-and-values.html
rickbutton: I've simplified this based on @littledan 's suggestion to just invoke %Array.prototype.join% for `TupleToString`.

spec/immutable-data-structures.html
rickbutton: Makes sense to me. The simplest solution here is to instead just copy over %Array.prototype.toLocaleString% in this case, which I've done here.

spec/abstract-operations.html
```diff
@@ -617,54 +617,76 @@ <h1>RequireObjectCoercible ( _argument_ )</h1>
...
+  </emu-clause>
+
+  <emu-clause id="sec-samevalue" aoid="SameValue">
+      <h1>SameValue ( _x_, _y_ )</h1>
+      <p>The abstract operation SameValue takes arguments _x_ (an ECMAScript language value) and _y_ (an ECMAScript language value) and returns a completion record whose [[Type]] is ~normal~ and whose [[Value]] is a Boolean. It performs the following steps when called:</p>
+      <emu-alg>
+        1. If Type(_x_) is different from Type(_y_), return *false*.
+        1. If Type(_x_) is <del>Number or BigInt</del><ins>Number, BigInt, Record, or Tuple</ins>, then
+          1. Return ! Type(_x_)::sameValue(_x_, _y_).
+        1. Return ! SameValueNonNumeric(_x_, _y_).
```
ljharb, line 629: SameValueNonNumeric seems like it’ll need a rename now

spec/data-types-and-values.html
```diff
@@ -9,27 +9,109 @@ <h1>ECMAScript Language Types</h1>
...
             1. Assert: Type(_argument_) is Record.
             1. Return *"[record]"*.
         </emu-alg>
       </emu-clause>
+
+      <emu-clause id="sec-immutable-types-record-equal">
+        <h1>Record::equal ( _x_, _y_, _elementEqual_ )</h1>
+        <p>The abstract operation Record::equal takes arguments _x_ (a Record), _y_ (a Record), and _elementEqual_ (a function object). _elementEqual_ will be invoked, with *undefined* as the receiver. It performs the following steps when called:</p>
+        <emu-alg>
+          1. Let _xList_ be a new List containing the same values as _x_.[[Fields]].
```
ljharb, line 25: Why a new List? Isn’t this a List already?

spec/data-types-and-values.html
```diff
@@ -9,27 +9,109 @@ <h1>ECMAScript Language Types</h1>
...
+          1. Let _yList_ be a new List containing the same values as _y_.[[Fields]].
+          1. Let _xLen_ be the length of List _xList_.
+          1. Let _yLen_ be the length of List _yList_.
+          1. If _xLen_ &ne; _yLen_, Return *false*.
```
ljharb, line 29:
```suggestion
          1. If _xLen_ &ne; _yLen_, return *false*.
```

spec/data-types-and-values.html
```diff
@@ -9,27 +9,109 @@ <h1>ECMAScript Language Types</h1>
...
+        <p>The abstract operation Record::toString takes argument _argument_ (a Record) and convertes _argument_ to a String. It performs the following steps when called:</p>
         <emu-alg>
             1. Assert: Type(_argument_) is Record.
             1. Return *"[record]"*.
         </emu-alg>
       </emu-clause>
+
+      <emu-clause id="sec-immutable-types-record-equal">
+        <h1>Record::equal ( _x_, _y_, _elementEqual_ )</h1>
+        <p>The abstract operation Record::equal takes arguments _x_ (a Record), _y_ (a Record), and _elementEqual_ (a function object). _elementEqual_ will be invoked, with *undefined* as the receiver. It performs the following steps when called:</p>
```
ljharb, line 23: It seems like elementEqual should be an abstract closure instead of a JS function. Is there a reason it’d ever be user code?

spec/abstract-operations.html
rickbutton: I agree, I couldn't think of a better name so I left it the same.

spec/data-types-and-values.html
rickbutton: nope, no reason to expose this to user code. abstract closure works for me.

spec/abstract-operations.html
rickbutton: SameValueNonNumericOrRecordOrTuple ? 🤣

ljharb: `SameValueNonGeneric`? since the things that are exempted from it are using the generic-ish T::foo() approach?

rickbutton:
I'm gonna merge this PR so that we can base more changes on it, and it looks to be in a good shape. Thanks all for the reviews, I am v grateful. Feel free to open a PR with additional suggestions/comments.

PR #153, opened 2020-07-12
Fix definition of tuple in glossary

### LINKED ISSUE #152: Fix naming error in Glossary
The glossary contains the following (incorrect) definition of a tuple:

> A new, deeply immutable, compound primitive type data structure, proposed in this document, that is analogous to Objects. #[1, 2, 3, 4]

This should be corrected to:
> A new, deeply immutable, compound primitive type data structure, proposed in this document, that is analogous to **Arrays**. #[1, 2, 3, 4] 

(emphasis mine)

mkeoliya:
Fixes #152.

### DIFF
--- a/README.md
+++ b/README.md
@@ -657,7 +657,7 @@ A new, deeply immutable, compound primitive type data structure, proposed in thi
 
 #### Tuple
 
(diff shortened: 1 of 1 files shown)


rricard:
Thank you! that was an oversight!

PR #157, opened 2020-07-15
Make Tuple.prototype.toString be Array.prototype.toString

rricard:


rricard:
Note: TupleToString in the previous version of the spec text was never defined

PR #158, opened 2020-07-15
simplification of Tuple.prototype.sorted

rickbutton:
I'm not entirely sure the right way to word this bit of prose while keeping the amount of duplication low. Suggestions welcome.

### DIFF
--- a/spec/immutable-data-structures.html
+++ b/spec/immutable-data-structures.html
@@ -327,7 +327,7 @@
           1. If _comparefn_ is not *undefined* and IsCallable(_comparefn_) is *false*, throw a *TypeError* exception.
           1. Let _T_ be ? thisTupleValue(*this* value).
           1. Let _list_ be a new List containing the elements of _T_.[[Sequence]].
-          1. Let _len_ be the length of list _list_.
+          1. Let _len_ be the length of _list_.
         </emu-alg>
         <p>A new Tuple containing the elements from the original Tuple is returned, where the elements are sorted. The <em>sort order</em> is the ordering, after completion of this function, of the elements of _list_. The <em>sort order</em> of the result of *sorted* is the same order as if the elements of _list_ were sorted in an Array via %Array.prototype.sort% with _comparefn_ as the first argument.</p>
       </emu-clause>
--- a/spec/immutable-data-structures.html
+++ b/spec/immutable-data-structures.html
@@ -318,59 +318,18 @@
           1. Return a new Tuple value whose [[Sequence]] is _newList_.
         </emu-alg>
       </emu-clause>
-      <emu-clause id="sec-array.prototype.sort">
+      <emu-clause id="sec-tuple.prototype.sorted">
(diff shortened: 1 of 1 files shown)


spec/immutable-data-structures.html
```diff
@@ -318,59 +318,18 @@ <h1>Tuple.prototype.sliced ( _start_, _end_ )</h1>
...
-        </emu-alg>
-        <p>The <em>sort order</em> is the ordering, after completion of this function, of the elements of _list_. The result of the `sort` function is then determined as follows:</p>
-        <p>If _comparefn_ is not *undefined* and is not a consistent comparison function for the elements of this array (see below), the sort order is implementation-defined. The sort order is also implementation-defined if _comparefn_ is *undefined* and SortCompare does not act as a consistent comparison function.</p>
-        <p>The following steps are taken:</p>
-        <emu-alg>
-          1. Perform an implementation-dependent sequence of operations (such as retrieving the element at a specific index, or mutating the value at a specific index) on _list_, such that:
-            * The arguments for calls to SortCompare are values in the list _list_.
-            * If an abrupt completion is returned from any of these operations, it is immediately returned as the value of this function.
-          1. Return a new Tuple value whose [[Sequence]] is _list_.
+          1. Let _len_ be the length of list _list_.
```
ljharb, line 330:
```suggestion
          1. Let _len_ be the length of _list_.
```

spec/immutable-data-structures.html
```diff
@@ -318,59 +318,18 @@ <h1>Tuple.prototype.sliced ( _start_, _end_ )</h1>
...
-            If _a_ &lt;<sub>CF</sub> _b_ and _b_ &lt;<sub>CF</sub> _c_, then _a_ &lt;<sub>CF</sub> _c_ (transitivity of &lt;<sub>CF</sub>)
-          </li>
-          <li>
-            If _a_ &gt;<sub>CF</sub> _b_ and _b_ &gt;<sub>CF</sub> _c_, then _a_ &gt;<sub>CF</sub> _c_ (transitivity of &gt;<sub>CF</sub>)
-          </li>
-        </ul>
-        <emu-note>
-          <p>The above conditions are necessary and sufficient to ensure that _comparefn_ divides the set _S_ into equivalence classes and that these equivalence classes are totally ordered.</p>
-        </emu-note>
+        <p>A new Tuple containing the elements from the original Tuple is returned, where the elements are sorted. The <em>sort order</em> is the ordering, after completion of this function, of the elements of _list_. The <em>sort order</em> of the result of *sorted* is the same order as if the elements of _list_ were sorted in an Array via %Array.prototype.sort% with _comparefn_ as the first argument.</p>
```
littledan, line 332: I feel like this should be an algorithm step rather than a paragraph, and talk about creating a Tuple whose [[Sequence]] is this sorted _list_, but this is also an unimportant nit, and it's fine to keep iterating here later.

rickbutton: I'll merge this in for now and make a note to take a second look at this later.

PR #159, opened 2020-07-15
refactored R/T specific abstract ops to not use Type::method format

rickbutton:


spec/data-types-and-values.html
```diff
@@ -10,17 +10,17 @@ <h1>The Record Type</h1>
...
             1. Assert: Type(_argument_) is Record.
             1. Return *"[object Record]"*.
         </emu-alg>
       </emu-clause>
 
       <emu-clause id="sec-immutable-types-record-equal">
-        <h1>Record::equal ( _x_, _y_, _elementEqual_ )</h1>
-        <p>The abstract operation Record::equal takes arguments _x_ (a Record), _y_ (a Record), and _elementEqual_ (a function object). _elementEqual_ will be invoked, with *undefined* as the receiver. It performs the following steps when called:</p>
+        <h1>RecordEqual ( _x_, _y_, _elementEqual_ )</h1>
+        <p>The abstract operation RecordEqual takes arguments _x_ (a Record), _y_ (a Record), and _elementEqual_ (a function object). _elementEqual_ will be invoked, with *undefined* as the receiver. It performs the following steps when called:</p>
```
littledan, line 23: I missed this before, but _elementEqual_ is actually an "abstract closure", not a function object, and should be identified as such, and called with the syntax `_elementEqual_(args)` rather than with Call.

spec/data-types-and-values.html
```diff
@@ -69,16 +69,16 @@ <h1>The Tuple Type</h1>
       <p>The Tuple type is the set of all finite and ordered sequences of ECMAScript primitive values including Record and Tuple. Each tuple value holds an associated [[Sequence]] List which is a list of primitive values. The [[Sequence]] List is integer-indexed. The [[Sequence]] List and its values are never modified.</p>
 
       <emu-clause id="sec-immutable-types-tuple-tostring">
-        <h1>Tuple::toString ( _argument_, [ _separator_ ] )</h1>
-        <p>The abstract operation Tuple::toString takes arguments _argument_ (a Tuple), and _separator_ (a String). It performs the following steps when called:</p>
+        <h1>TupleToString ( _argument_, [ _separator_ ] )</h1>
```
littledan, line 72: This algorithm is never called with a separator. Probably it shouldn't take that as a parameter, and just call Array.prototype.join with no arguments.

PR #160, opened 2020-07-16
Add missing ToIndex conversion in Tuple.prototype.with

nicolo-ribaudo:
Without first ensuring that the index is a number, `>=` isn't defined.

### DIFF
--- a/spec/immutable-data-structures.html
+++ b/spec/immutable-data-structures.html
@@ -615,9 +615,10 @@
           1. Let _T_ be ? thisTupleValue(*this* value).
           1. Let _list_ be a new List containing the elements of _T_.[[Sequence]].
           1. Let _length_ be the length of list _list_.
-          1. If _index_ &lt; 0 or _index_ >= _length_, throw a *RangeError* exception.
+          1. Let _I_ be ? ToIndex(_index_).
+          1. If _I_ &ge; _length_, throw a *RangeError* exception.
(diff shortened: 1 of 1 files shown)


spec/immutable-data-structures.html
```diff
@@ -615,9 +615,10 @@ <h1>Tuple.prototype.with ( _index_, _value_ )</h1>
           1. Let _T_ be ? thisTupleValue(*this* value).
           1. Let _list_ be a new List containing the elements of _T_.[[Sequence]].
           1. Let _length_ be the length of list _list_.
-          1. If _index_ &lt; 0 or _index_ >= _length_, throw a *RangeError* exception.
+          1. Let _I_ be ? ToIndex(_index_).
+          1. If _I_ &ge; _length_, throw a *RangeError* exception.
```
ljharb, line 619: Should this not wrap around, like .item?

littledan: IMO we should stay as similar to Arrays and TypedArrays as possible. We would also have a `Tuple.prototype.item` method, if that proposal goes through.

nicolo-ribaudo: I would really like to see `.with()` working with negative indexes, I'll open an issue to discuss about it.

I'd prefer to keep this PR with the `RangeError`, since this is just fixing a spec bug and not changing the existing functionality.

PR #162, opened 2020-07-16
Write spec for JSON.parseImmutable

nicolo-ribaudo:
This pull request adds the definition of the `JSON.parseImmutable` function.
The implementation proposed here passes `undefined` as the `this` object of the `reviver` function, for the reason explained at https://github.com/tc39/proposal-record-tuple/issues/120#issuecomment-658997804, but we can still change it in the future.

### DIFF
--- a/spec/structured-data.html
+++ b/spec/structured-data.html
@@ -1,0 +1,111 @@
+<!DOCTYPE html>
+<meta charset="utf8" />
+
+<emu-clause id="sec-structured-data">
+  <h1>Structured Data</h1>
+
+  <emu-clause id="sec-json-object">
+    <h1>The JSON Object</h1>
+
+    <emu-clause id="sec-json.parse">
+      <h1>JSON.parse ( _text_ [ , _reviver_ ] )</h1>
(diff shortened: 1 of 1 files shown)


spec/structured-data.html
```diff
@@ -0,0 +1,111 @@
...
+              1. Let _fields_ be a new empty List.
+              1. For each element _prop_ of _props_,
+                1. Let _childName_ be ! GetV(_prop_, `"0"`).
+                1. Let _childValue_ be ! GetV(_prop_, `"1"`).
+                1. Let _newElement_ be ? BuildImmutableProperty(_childValue_, _childName_, _reviver_).
+                1. Assert: Type(_newElement_) is not Object.
+                1. If _newElement_ is not *undefined*, then:
+                  1. Let _field_ be the Record { [[Key]]: _childName_, [[Value]]: _newElement_ }.
+                  1. Append _field_ to _fields_.
+              1. Let _immutable_ be a new Record value whose [[Fields]] is _fields_.
```
littledan, line 88: Hmm, the fields here might be unsorted. We should probably make an abstract operation for deduplicating, sorting and making a Record. I think actually there used to be an operation for this, but then I told @rricard to inline it, but that was a bad suggestion :) This is also needed for Record.fromEntries and Record(), which also fail to sort the keys. (I guess just record literals and `fromEntries` actually need the deduplication, but it doesn't hurt to include that path for all sources of creating new Records.)

PR #164, opened 2020-07-16
Remove mentions of static Record methods

nicolo-ribaudo:
These methods are not in the spec, and they can be replaced with `Object.*` functions.

### DIFF
--- a/NS-Proto-Appendix.md
+++ b/NS-Proto-Appendix.md
@@ -236,12 +236,12 @@ Returns a `Tuple` with the same elements as the original `Tuple` except the elem
 - `Object.is()` - When comparing records and tuples it behaves exactly like `===`
 - `Object.entries()`, `Object.keys()`, `Object.values()` - The resulting arrays are sorted alphabetically by key
 - `Object.assign()` - Copy properties from a record or tuple to another object. **Caveat**: The first argument cannot be a record or a tuple, because they are immutable
-- `Object.isExtensible()`, `Object.isFrozen()`, `Object.isSealed()` - They always return `false` for records and tuples
+- `Object.isExtensible()`, `Object.isFrozen()`, `Object.isSealed()` - When called on records and tuples, they always return `false`, `true`, and `true` respectively
 - `Object.preventExtension()`, `Object.freeze()`, `Object.seal()` - Records and tuples are already immutable, so these methods will not throw but won't have any effect
 - `Object.getOwnPropertyDescriptor()`, `Object.getOwnPropertyDescriptors()` - The resulting descriptors all have `writable: false, enumerable: true, configurable: false`
 - `Object.getOwnPropertyNames()`, `Object.getOwnPropertySymbols()`, `Object.getPrototypeOf()` - Behave similarly to when called on objects
 
-- `Object.prototype.toString`, `Object.prototype.toLocaleString` - Return `[object Record]` and `[object Tuple]`
+- `Object.prototype.toString`, `Object.prototype.toLocaleString` - Return `[object Record]` and `[object Tuple]` when a Record or Tuple is the receiver
 - `Object.prototype.valueOf` - Boxes a primitive record or tuple into an object
 - `Object.prototype.propertyIsEnumerable` - Always returns `true` for existing properties
 - `Object.prototype.hasOwnProperty`
@@ -249,5 +249,5 @@ Returns a `Tuple` with the same elements as the original `Tuple` except the elem
 
 ## `Object` methods which don't work with `Record` and `Tuple`
 
-- `Object.create()` - Records and tuples cannot have a prototype
-- `Object.defineProperties()`, `Object.defineProperty()`, `Object.setPrototypeOf()` - Records and tuples are non-observable, not writable by nature
+- `Object.create()` - Records and tuples cannot be used as prototypes, unless you wrap them in objects (`Object(#{})`)
+- `Object.defineProperties()`, `Object.defineProperty()`, `Object.setPrototypeOf()` - Records and tuples are immutable by design.
--- a/NS-Proto-Appendix.md
+++ b/NS-Proto-Appendix.md
@@ -18,10 +18,10 @@ Similar to `Object.fromEntries`, but created a new `Record`.
 - `Object.entries()`
 - `Object.keys()`
 - `Object.values()`
+- `Object.assign()` - **Caveat**: The first argument cannot be a record or a tuple, because they are immutable
 
 ## `Object` namespace functions not working with `Record`
 
-- `Object.assign()` - You can't assign new properties to a record.
 - `Object.create()` - `Record` has no prototype so `Record.create()` would not have any meaning.
 - `Object.defineProperties()`, `Object.defineProperty()` - A `Record` is non-observable, not writable by nature
 - `Object.freeze()`, `Object.seal()` - `Record` is not mutable
--- a/NS-Proto-Appendix.md
+++ b/NS-Proto-Appendix.md
@@ -8,41 +8,27 @@ Converts shallowly an object to a record. If the object has a non-const value, a
 
 Checks whether the parameter is either a Record primitive or Record wrapper.
 
-## `Record.assign(...args: Record[]) -> Record`
-
-Merges all records passed as arguments into one **new** record returned by the function. The latest argument will override arguments before when merging.
-
-## `Record.entries(record: Record) -> Tuple`
-
-Returns all of the entries in a Record as a Tuple of 2-values Tuples (key, value) in the same order as `Record.keys()`.
-
 ## `Record.fromEntries(iterator: Iterator): Record`
 
-Takes an iterator of 2-element array-like values and creates a record out of it. If the iterator returns something other than an array-like tuple, a TypeError will be raised.
+Similar to `Object.fromEntries`, but created a new `Record`.
 
-## `Record.keys(record: Record) -> Tuple`
+## `Object` namespace functions working with `Record`
 
-Returns a Tuple containing the values serving as keys. Order will be the one defined in the main proposal document.
+- `Object.is()`
+- `Object.entries()`
+- `Object.keys()`
+- `Object.values()`
 
-## `Record.toString(record: Record) -> String`
+## `Object` namespace functions not working with `Record`
 
-To be defined - it can actually show the full value.
-
-## `Record.values(record: Record) -> Tuple`
-
-Returns all of the values in a Record in the same order as the associated keys in `Record.keys()`.
-
-## `Object` namespace functions not available in `Record`
-
+- `Object.assign()` - You can't assign new properties to a record.
 - `Object.create()` - `Record` has no prototype so `Record.create()` would not have any meaning.
 - `Object.defineProperties()`, `Object.defineProperty()` - A `Record` is non-observable, not writable by nature
 - `Object.freeze()`, `Object.seal()` - `Record` is not mutable
 - `Object.getOwnPropertyDescriptor()`, `Object.getOwnPropertyDescriptors()` - Does not make sense as there is no `defineProperty`
 - `Object.getOwnPropertyNames()`, `Object.getOwnPropertySymbols()`, `Object.getPrototypeOf()`, `Object.setPrototypeOf()` - Does not make sense without a prototype
-- `Object.is()` - `Record.is()` would be exactly the same as `Object.is()`, we can eventually make it an alias
 - `Object.isExtensible()`, `Object.isFrozen()`, `Object.isSealed()` - `Record` is not mutable anyway
-- `Object.hasOwnProperty()`, `Object.isPrototypeOf()`, `Object.isEnumerable()` - As we can't set those things up as seen previously, those are not available on `Record`
-- `Object.toLocaleString()`, `Object.valueOf()` - Only used by objects overloading Object, we can't overload a Record
+- `Object.isPrototypeOf()`, `Object.isEnumerable()` - As we can't set those things up as seen previously, those are not available on `Record`
 
 # `Tuple` namespace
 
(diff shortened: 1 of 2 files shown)


NS-Proto-Appendix.md
```diff
@@ -8,41 +8,27 @@ Converts shallowly an object to a record. If the object has a non-const value, a
...
+- `Object.assign()` - You can't assign new properties to a record.
 - `Object.create()` - `Record` has no prototype so `Record.create()` would not have any meaning.
 - `Object.defineProperties()`, `Object.defineProperty()` - A `Record` is non-observable, not writable by nature
 - `Object.freeze()`, `Object.seal()` - `Record` is not mutable
 - `Object.getOwnPropertyDescriptor()`, `Object.getOwnPropertyDescriptors()` - Does not make sense as there is no `defineProperty`
 - `Object.getOwnPropertyNames()`, `Object.getOwnPropertySymbols()`, `Object.getPrototypeOf()`, `Object.setPrototypeOf()` - Does not make sense without a prototype
-- `Object.is()` - `Record.is()` would be exactly the same as `Object.is()`, we can eventually make it an alias
 - `Object.isExtensible()`, `Object.isFrozen()`, `Object.isSealed()` - `Record` is not mutable anyway
-- `Object.hasOwnProperty()`, `Object.isPrototypeOf()`, `Object.isEnumerable()` - As we can't set those things up as seen previously, those are not available on `Record`
-- `Object.toLocaleString()`, `Object.valueOf()` - Only used by objects overloading Object, we can't overload a Record
```
nicolo-ribaudo, line 45: Those aren't properties of `Object`, they come from `Object.__proto__ === Function.prototype` so they will also be present on `Record`.

NS-Proto-Appendix.md
```diff
@@ -8,41 +8,27 @@ Converts shallowly an object to a record. If the object has a non-const value, a
...
-Returns all of the values in a Record in the same order as the associated keys in `Record.keys()`.
+- `Object.is()`
+- `Object.entries()`
+- `Object.keys()`
+- `Object.values()`
 
-## `Object` namespace functions not available in `Record`
+## `Object` namespace functions not working with `Record`
 
+- `Object.assign()` - You can't assign new properties to a record.
```
ljharb, line 24: However, you should be able to pass records or Tuple as sources, no?

nicolo-ribaudo: Right, I will update the desc. The spec should already work, because it calls `ToObject` on the arguments.

NS-Proto-Appendix.md
```diff
@@ -8,41 +8,27 @@ Converts shallowly an object to a record. If the object has a non-const value, a
...
 
-## `Record.values(record: Record) -> Tuple`
+## `Object` namespace functions working with `Record`
 
-Returns all of the values in a Record in the same order as the associated keys in `Record.keys()`.
+- `Object.is()`
+- `Object.entries()`
+- `Object.keys()`
+- `Object.values()`
+- `Object.assign()` - **Caveat**: The first argument cannot be a record or a tuple, because they are immutable
```
littledan, line 21: As written, on a skim, it is not clear that the caveat only applies to the last one. I would include (maybe in a follow-on PR) short descriptions of the behavior that comes from each of these. That will make it easy to say that you're only referring to the second argument of `Object.assign`

NS-Proto-Appendix.md
```diff
@@ -8,41 +8,27 @@ Converts shallowly an object to a record. If the object has a non-const value, a
...
 
-## `Object` namespace functions not available in `Record`
+## `Object` namespace functions not working with `Record`
```
littledan, line 23: English nit: "...functions which don't work with"

NS-Proto-Appendix.md
```diff
@@ -8,41 +8,27 @@ Converts shallowly an object to a record. If the object has a non-const value, a
...
 
 - `Object.create()` - `Record` has no prototype so `Record.create()` would not have any meaning.
 - `Object.defineProperties()`, `Object.defineProperty()` - A `Record` is non-observable, not writable by nature
 - `Object.freeze()`, `Object.seal()` - `Record` is not mutable
 - `Object.getOwnPropertyDescriptor()`, `Object.getOwnPropertyDescriptors()` - Does not make sense as there is no `defineProperty`
 - `Object.getOwnPropertyNames()`, `Object.getOwnPropertySymbols()`, `Object.getPrototypeOf()`, `Object.setPrototypeOf()` - Does not make sense without a prototype
-- `Object.is()` - `Record.is()` would be exactly the same as `Object.is()`, we can eventually make it an alias
 - `Object.isExtensible()`, `Object.isFrozen()`, `Object.isSealed()` - `Record` is not mutable anyway
```
littledan, line 30: These will "just work" and return the right answer without spec changes.

NS-Proto-Appendix.md
```diff
@@ -8,41 +8,27 @@ Converts shallowly an object to a record. If the object has a non-const value, a
...
+- `Object.keys()`
+- `Object.values()`
+- `Object.assign()` - **Caveat**: The first argument cannot be a record or a tuple, because they are immutable
 
-## `Object` namespace functions not available in `Record`
+## `Object` namespace functions not working with `Record`
 
 - `Object.create()` - `Record` has no prototype so `Record.create()` would not have any meaning.
 - `Object.defineProperties()`, `Object.defineProperty()` - A `Record` is non-observable, not writable by nature
 - `Object.freeze()`, `Object.seal()` - `Record` is not mutable
```
littledan, line 27: It's fine to call these functions; they won't do anything, since Records and Tuples are already frozen/sealed.

NS-Proto-Appendix.md
```diff
@@ -8,41 +8,27 @@ Converts shallowly an object to a record. If the object has a non-const value, a
...
+- `Object.entries()`
+- `Object.keys()`
+- `Object.values()`
+- `Object.assign()` - **Caveat**: The first argument cannot be a record or a tuple, because they are immutable
 
-## `Object` namespace functions not available in `Record`
+## `Object` namespace functions not working with `Record`
 
 - `Object.create()` - `Record` has no prototype so `Record.create()` would not have any meaning.
 - `Object.defineProperties()`, `Object.defineProperty()` - A `Record` is non-observable, not writable by nature
```
littledan, line 26: This function will "work" on a Record wrapper but not do anything really useful (since they act like they're frozen). It'll throw on Record and Tuple values, which is probably fine to leave as is.

An alternative for Object.defineProperty would be to make it call ToObject on Records and Tuples, and do [[DefineProperty]] on that. I don't have much of an opinion about whether we do this now; it maybe more of an attractive change if there is a more general future value types proposal.

NS-Proto-Appendix.md
```diff
@@ -8,41 +8,27 @@ Converts shallowly an object to a record. If the object has a non-const value, a
...
 - `Object.freeze()`, `Object.seal()` - `Record` is not mutable
 - `Object.getOwnPropertyDescriptor()`, `Object.getOwnPropertyDescriptors()` - Does not make sense as there is no `defineProperty`
```
littledan, line 28: This will just work.

NS-Proto-Appendix.md
```diff
@@ -8,41 +8,27 @@ Converts shallowly an object to a record. If the object has a non-const value, a
...
 - `Object.getOwnPropertyNames()`, `Object.getOwnPropertySymbols()`, `Object.getPrototypeOf()`, `Object.setPrototypeOf()` - Does not make sense without a prototype
```
littledan, line 29: `Object.getOwnPropertyNames`, `Object.getOwnPropertySymbols` and `Object.getPrototypeOf` will just work.

`Object.setPrototypeOf` will throw, and my analysis about this matches that of `Object.defineProperty` described above.

NS-Proto-Appendix.md
```diff
@@ -8,41 +8,27 @@ Converts shallowly an object to a record. If the object has a non-const value, a
...
-- `Object.is()` - `Record.is()` would be exactly the same as `Object.is()`, we can eventually make it an alias
 - `Object.isExtensible()`, `Object.isFrozen()`, `Object.isSealed()` - `Record` is not mutable anyway
-- `Object.hasOwnProperty()`, `Object.isPrototypeOf()`, `Object.isEnumerable()` - As we can't set those things up as seen previously, those are not available on `Record`
-- `Object.toLocaleString()`, `Object.valueOf()` - Only used by objects overloading Object, we can't overload a Record
+- `Object.isPrototypeOf()`, `Object.isEnumerable()` - As we can't set those things up as seen previously, those are not available on `Record`
```
littledan, line 31: This is `Object.prototype.isPrototypeOf`, which will return `false` when called on a Record or Tuple, but `true` when called on a wrapper, comparing with the wrapper prototype. Similarly to `setPrototypeOf` and `defineProperty`, we might want to revisit this in a more generalized context, but conservatively returning `false` seems fine for now.

`Object.prototype.propertyIsEnumerable` should just work, since it starts with ToObject.

These would need to be invoked with `.call`, since `Object.prototype` is not in the prototype chain of Record wrappers. But you can call them directly on Tuples.

NS-Proto-Appendix.md
littledan: I guess you mean these are on Object.prototype. These would "just work" when `.call`'ed, but not be found on the prototype chain for Records.

NS-Proto-Appendix.md
ljharb: It won’t throw if you try to define the same values.

NS-Proto-Appendix.md
ljharb: also gOPD makes sense even if you never define anything :-)

nicolo-ribaudo:
I have left `defineProperty` and `setPrototypeOf` as not working for now

NS-Proto-Appendix.md
```diff
@@ -258,3 +226,28 @@ Returns a `Tuple Iterator` object that contains the values for each index in the
...
+# Compatibility with the existing `Object` methods
+
+> NOTE: Since `Record` doesn't inherit from `Object`, you can't directly call the `Object.prototype.*` methods but you have to use `.call()`.
+
+## `Object` methods which work with `Record` and `Tuple`
+
+- `Object.is()` - When comparing records and tuples it behaves exactly like `===`
+- `Object.entries()`, `Object.keys()`, `Object.values()` - The resulting arrays are sorted alphabetically by key
+- `Object.assign()` - Copy properties from a record or tuple to another object. **Caveat**: The first argument cannot be a record or a tuple, because they are immutable
+- `Object.isExtensible()`, `Object.isFrozen()`, `Object.isSealed()` - They always return `false` for records and tuples
```
ljharb, line 239: they return false, true, true, respectively.

NS-Proto-Appendix.md
```diff
@@ -258,3 +226,28 @@ Returns a `Tuple Iterator` object that contains the values for each index in the
...
+- `Object.preventExtension()`, `Object.freeze()`, `Object.seal()` - Records and tuples are already immutable, so these methods will not throw but won't have any effect
+- `Object.getOwnPropertyDescriptor()`, `Object.getOwnPropertyDescriptors()` - The resulting descriptors all have `writable: false, enumerable: true, configurable: false`
+- `Object.getOwnPropertyNames()`, `Object.getOwnPropertySymbols()`, `Object.getPrototypeOf()` - Behave similarly to when called on objects
+
+- `Object.prototype.toString`, `Object.prototype.toLocaleString` - Return `[object Record]` and `[object Tuple]`
```
ljharb, line 244:
```suggestion
- `Object.prototype.toString`, `Object.prototype.toLocaleString` - Return `[object Record]` and `[object Tuple]` when a Record or Tuple is the receiver
```

NS-Proto-Appendix.md
```diff
@@ -258,3 +226,28 @@ Returns a `Tuple Iterator` object that contains the values for each index in the
...
+- `Object.prototype.valueOf` - Boxes a primitive record or tuple into an object
+- `Object.prototype.propertyIsEnumerable` - Always returns `true` for existing properties
+- `Object.prototype.hasOwnProperty`
+- `Object.prototype.isPrototypeOf`
+
+## `Object` methods which don't work with `Record` and `Tuple`
+
+- `Object.create()` - Records and tuples cannot have a prototype
```
ljharb, line 252: they could *be* a [[Prototype]], though, so I’m not sure how this entry makes sense - you certainly can’t produce a non-Object from an Object method, but you can pass a Record or Tuple into Object.create anywhere?

NS-Proto-Appendix.md
```diff
@@ -258,3 +226,28 @@ Returns a `Tuple Iterator` object that contains the values for each index in the
...
+- `Object.defineProperties()`, `Object.defineProperty()`, `Object.setPrototypeOf()` - Records and tuples are non-observable, not writable by nature
```
ljharb, line 253: non-observable doesn’t seem like the right term here

NS-Proto-Appendix.md
littledan: It will, because `Object.defineProperty` throws on non-objects, in its first step, before we get to the [[DefineProperty]] behavior for Record wrappers. So, that's something that we can go either way on relaxing IMO. We can also go either way on relaxing it for `Reflect`. Ultimately, you can always access this functionality by casting to Object explicitly, and also the functionality is not very useful.

ljharb: ah, my mistake for the primitive form. seems reasonable to keep the exception for the primitive if there’s no useful defining to be done in it.

NS-Proto-Appendix.md
littledan: Well, the current spec text would throw a TypeError. You'd have to do `Object.create(Object(#{}))` explicitly. I could go either way on whether we relax this constraint. (I do agree that the explanation doesn't make sense.)

NS-Proto-Appendix.md
nicolo-ribaudo: I'm not sure how I missed it 😅

PR #165, opened 2020-07-17
@@isConcatSpreadable for Tuples

rickbutton:


littledan:
IMO we should follow Arrays and update the [IsConcatSpreadable](https://tc39.es/ecma262/#sec-isconcatspreadable) algorithm directly instead of defining this property, while still respecting that property if someone does define it on `Tuple.prototype`.

ljharb:
good call

spec/abstract-operations.html
```diff
@@ -744,4 +744,17 @@ <h1>AddValueToTupleSequenceList ( _sequence_, _value_ )</h1>
...
   </emu-clause>
+
+  <emu-clause id="sec-isconcatspreadable" aoid="IsConcatSpreadable">
+      <h1>Runtime Semantics: IsConcatSpreadable ( _O_ )</h1>
+      <p>The abstract operation IsConcatSpreadable takes argument _O_. It performs the following steps when called:</p>
+      <emu-alg>
+        1. If Type(_O_) is not Object, return *false*.
+        1. Let _spreadable_ be ? Get(_O_, @@isConcatSpreadable).
+        1. If _spreadable_ is not *undefined*, return ! ToBoolean(_spreadable_).
+        1. <ins>If _O_ has a [[TupleData]] internal slot, return *true*</ins>
```
littledan, line 755: If Type(_O_) is Tuple or if Type(_O_) is Object and _O_ has a [[TupleData]] internal slot, return *true*

spec/abstract-operations.html
```diff
@@ -744,4 +744,17 @@ <h1>AddValueToTupleSequenceList ( _sequence_, _value_ )</h1>
       </emu-alg>
     </emu-clause>
   </emu-clause>
+
+  <emu-clause id="sec-isconcatspreadable" aoid="IsConcatSpreadable">
+      <h1>Runtime Semantics: IsConcatSpreadable ( _O_ )</h1>
+      <p>The abstract operation IsConcatSpreadable takes argument _O_. It performs the following steps when called:</p>
+      <emu-alg>
+        1. If Type(_O_) is not Object, return *false*.
```
littledan, line 752: Oops, we already returned false up here! "is not Object or Tuple", probably this should say, with a GetV below.

rickbutton:
@littledan apologies, I iterated on this PR a bunch of times in the middle of your review. Making tuples -always- concat spreadable was a mistake, I moved the check to after the symbol lookup. I think it would be weird if Tuples were the only thing that was forced into concat-spreadable with no opt opt.

spec/abstract-operations.html
```diff
@@ -744,4 +744,17 @@ <h1>AddValueToTupleSequenceList ( _sequence_, _value_ )</h1>
...
   </emu-clause>
+
+  <emu-clause id="sec-isconcatspreadable" aoid="IsConcatSpreadable">
+      <h1>Runtime Semantics: IsConcatSpreadable ( _O_ )</h1>
+      <p>The abstract operation IsConcatSpreadable takes argument _O_. It performs the following steps when called:</p>
+      <emu-alg>
+        1. If Type(_O_) is not Object or Tuple, return *false*
+        1. Let _spreadable_ be ? Get<ins>V</ins>(_O_, @@isConcatSpreadable).
+        1. If _spreadable_ is not *undefined*, return ! ToBoolean(_spreadable_).
+        1. <ins>If Type(_O_) is Tuple or Type(_O_) is Object and _O_ has a [[TupleData]] internal slot, return *true*</ins>
```
ljharb, line 755: is there no internal `isTuple` AO for this?

rickbutton: not currently, we are manually checking for `TupleData` in the few places we need to. We could certainly add one.

ljharb: I'd prefer that for spec clarity :-)

rickbutton: sure thing, just added, also switched `Tuple.isTuple` over to using it.

PR #167, opened 2020-07-17
Broke off some things into a separate rationale section

rickbutton:
I also tweaked the wording for the "boxing objects" section, we haven't updated that section since iterating on the problem space via Symbols as WeakMap keys.


rickbutton:
gonna merge now, will iterate further in more PRs

PR #178, opened 2020-07-23
Fix wrong variable name declaration

telmen:


rricard:
Thanks for the correction, that being said, I'd like to keep the name `measure` and have it referenced afterwards instead of using `document`

telmen:
Closing in favor of #180

PR #190, opened 2020-07-28
Fixing example

jakearchibald:
The previous example used `document`, which is confusing due to `globalThis.document` in web pages, especially since `document.title` also exists.

The example also incorrectly referenced `document` when it intended to reference `document2`.


jakearchibald:
Closing in favour of https://github.com/tc39/proposal-record-tuple/pull/182

PR #202, opened 2020-09-27
Update README.md to fix typo in TS type

N2D4:
The `readonly` keyword is only allowed on TS tuples and TS arrays, not on other types (such as `Record`). Instead, there is the utility type `Readonly<T>`, which marks all properties as readonly.

[TS Playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBATgU1HAJgRgFzwQQxeAGwE8YAlJEVAHmjgEswBzAGhjAFcBbAIwTgD4YAXhgBvGDAD0kmAHJEeQkVkwoRAA4IYnECjoAzOnxh0IMJTE1xOdKFAQpzYGDjhwcJHGEdR26gloEtnw4BKoaCBAAdAAUaADMAKwALACUAFAS+iAgWGjMmTDcrlgATAUAvgDc6emgkLCIyCilWOSKYMRU5M00UPRMrBw8fPyCIuJSMmAgMHxwlBAAhIXZuTD5hcVwZZU1QA)


rickbutton:
closing this in favor of #299 , which is the same change but against the latest spec. Good catch!

PR #236, opened 2021-07-12
Only get enumerable properties when spreading in records

nicolo-ribaudo:
This aligns it with spread in object literals.

### DIFF
--- a/spec/expression.html
+++ b/spec/expression.html
@@ -119,8 +119,10 @@
           1. Let _from_ be ! ToObject(_source_).
           1. Let _keys_ be ? _from_.[[OwnPropertyKeys]]().
           1. For each element _nextKey_ of _keys_ in List order, do
-            1. Let _value_ be _from_.[[Get]](_nextKey_).
-            1. Perform ? AddPropertyIntoRecordEntriesList(_entries_, _nextKey_, _value_).
+            1. Let _desc_ be ? _from_.[[GetOwnProperty]](_nextKey_).
+            1. If _desc_ is not *undefined* and _desc_.[[Enumerable]] is *true*, then
+              1. Let _value_ be ? Get(_from_, _nextKey_).
+              1. Perform ? AddPropertyIntoRecordEntriesList(_entries_, _nextKey_, _value_).
           1. Return _entries_.
         </emu-alg>
         <emu-grammar>RecordPropertyDefinition : IdentifierReference</emu-grammar>


spec/expression.html
```diff
@@ -119,8 +119,10 @@ <h1>Runtime Semantics: RecordPropertyDefinitionEvaluation</h1>
           1. Let _from_ be ! ToObject(_source_).
           1. Let _keys_ be ? _from_.[[OwnPropertyKeys]]().
           1. For each element _nextKey_ of _keys_ in List order, do
-            1. Let _value_ be _from_.[[Get]](_nextKey_).
```
nicolo-ribaudo, line 122: This was another bug: `[[Get]]` expects two parameters. I replaced it with a call to `Get()`, which seems to be used more often in the spec when the receiver matches the object.

rricard (approved):

spec/expression.html
```diff
@@ -119,8 +119,10 @@ <h1>Runtime Semantics: RecordPropertyDefinitionEvaluation</h1>
           1. Let _from_ be ! ToObject(_source_).
           1. Let _keys_ be ? _from_.[[OwnPropertyKeys]]().
           1. For each element _nextKey_ of _keys_ in List order, do
-            1. Let _value_ be _from_.[[Get]](_nextKey_).
-            1. Perform ? AddPropertyIntoRecordEntriesList(_entries_, _nextKey_, _value_).
+            1. Let _desc_ be ? _from_.[[GetOwnProperty]](_nextKey_).
+            1. If _desc_ is not *undefined* and _desc_.[[Enumerable]] is *true*, then
```
jridgewell, line 123: Can it have non–enumerable keys?

acutmore: I think so yes, we could be spreading any object here.

```js
let x = { a: 1 };
Object.defineProperty(x, 'b', { value: 2});
let r = #{ ...x }; // #{ a: 1 };
```

ljharb: also `#{ ...[] }` has a non-enumerable `length` property, eg

spec/expression.html
```diff
@@ -119,8 +119,10 @@ <h1>Runtime Semantics: RecordPropertyDefinitionEvaluation</h1>
           1. Let _from_ be ! ToObject(_source_).
           1. Let _keys_ be ? _from_.[[OwnPropertyKeys]]().
           1. For each element _nextKey_ of _keys_ in List order, do
-            1. Let _value_ be _from_.[[Get]](_nextKey_).
-            1. Perform ? AddPropertyIntoRecordEntriesList(_entries_, _nextKey_, _value_).
+            1. Let _desc_ be ? _from_.[[GetOwnProperty]](_nextKey_).
+            1. If _desc_ is not *undefined* and _desc_.[[Enumerable]] is *true*, then
+              1. Let _value_ be ? Get(_from_, _nextKey_).
```
littledan, line 124: This looks "correct" in terms of matching https://tc39.es/ecma262/#sec-copydataproperties but I am a bit surprised that both cases don't use desc.[[Value]]--I guess to permit own getters? Still surprising that we repeat the MOP operation.

littledan (approved):

spec/expression.html
nicolo-ribaudo: Yes, spread invokes getters. I think the double get (GetOwnProperty + Get) might be observable via proxies, but if we want to change it it should be first done in the object case (and then copied in this proposal).

littledan: I agree. I think this PR is good to land.

PR #239, opened 2021-07-13
Fix different spec bugs

nicolo-ribaudo:


ljharb:
Why hardcoded Tuple mechanics instead of using IsConcatSpreadable? (which you’d get for free with the symbol on Tuple.prototype)

nicolo-ribaudo:
Mostly because the current paths after `IsConcatSpreadable(_x_)` assume that `_x_` is an object, which isn't true for tuples.

ljharb:
Given that i can `Number.prototype[Symbol.isConcatSpreadable] = true` or similar, that seems like a spec bug if that's the case - where in https://tc39.es/ecma262/#sec-array.prototype.concat is that assumption?

nicolo-ribaudo:
@ljharb `LengthOfArrayLike` (called if `spreadable` is true) starts with "Assert: Type(obj) is Object". If that's indeed a bug I can open a PR in the main spec.

nicolo-ribaudo:
Oh well, we are both wrong. Even if `Number.prototype[Symbol.isConcatSpreadable] = true`, `IsConcatSpreadable` currently returns `false` because it only returns `true` for _objects_ with a `[Symbol.isConcatSpreadable] = true` property.

If I make `IsConcatSpreadable` return `true` also for tuples, I would have to change that path which _currently_ can safely assume that the spreadable element is an object. (it might be worth to add an assertion to make it more clear?)

ljharb:
ah yes, that's true - i see step 1 in https://tc39.es/ecma262/#sec-isconcatspreadable.

I would expect that this proposal does indeed change most of the places requiring an Object to permit either Object, Record, or Tuple, including IsConcatSpreadable and LengthOfArrayLike.

nicolo-ribaudo:
I reverted the workaround for tuples in `.concat` but I didn't fix it yet. We need to go through the full spec for that, and I prefer to keep this PR slimmer and easily reviewable.

littledan:
Tuples already are concat-spreadable via https://tc39.es/proposal-record-tuple/#sec-isconcatspreadable so I am having trouble understanding the above discussion. It's true that a lot of paths need updating for non-objects, but this is a broader issue with the specification.

littledan (approved):
All great fixes!

PR #246, opened 2021-08-09
Throw when using symbol keys in the `Record` constructor

nicolo-ribaudo:
Fixes https://github.com/tc39/proposal-record-tuple/issues/237

### DIFF
--- a/spec/immutable-data-structures.html
+++ b/spec/immutable-data-structures.html
@@ -29,7 +29,7 @@
               1. If Type(_key_) is Symbol, throw a *TypeError* exception.
               1. Let _value_ be ? Get(_obj_, _key_).
               1. If Type(_value_) is Object, throw a *TypeError* exception.
-              1. Let _field_ be the Record { [[Key]]: _name_, [[Value]]: _value_ }.
+              1. Let _field_ be the Record { [[Key]]: _key_, [[Value]]: _value_ }.
               1. Append _field_ to the end of list _fields_.
           1. Return ! CreateRecord(_fields_).
         </emu-alg>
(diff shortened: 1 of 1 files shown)


rickbutton (approved):

spec/immutable-data-structures.html
```diff
@@ -21,14 +21,16 @@ <h1>Record ( _arg_ )</h1>
...
-            1. If Type(_value_) is Object, throw a *TypeError* exception.
-            1. Let _field_ be the Record { [[Key]]: _name_, [[Value]]: _value_ }.
-            1. Append _field_ to the end of list _fields_.
+          1. For each element _key_ of _keys_,
+            1. Let _desc_ be ? _obj_.[[GetOwnProperty]](_key_).
+            1. If _desc_ is not *undefined* and _desc_.[[Enumerable]] is *true*, then
+              1. If Type(_key_) is Symbol, throw a *TypeError* exception.
+              1. Let _value_ be ? Get(_obj_, _key_).
+              1. If Type(_value_) is Object, throw a *TypeError* exception.
+              1. Let _field_ be the Record { [[Key]]: _name_, [[Value]]: _value_ }.
```
rickbutton, line 32: `_name_` is undefined here, should this be `_key_`?

rickbutton (changes requested):

spec/immutable-data-structures.html
nicolo-ribaudo: Yes!
```suggestion
              1. Let _field_ be the Record { [[Key]]: _key_, [[Value]]: _value_ }.
```

rickbutton (approved):

PR #276, opened 2021-12-17
Update tuto & cookbook

rricard:
Removes boxes

Removes pushed

### DIFF
--- a/README.md
+++ b/README.md
@@ -806,7 +806,7 @@ A new, deeply immutable, compound primitive type data structure, proposed in thi
 
 #### Compound primitive types
 
-Values which act like other JavaScript primitives, but are composed of other constituent values. This document proposes the first three compound primitive types: `Record` and `Tuple`.
+Values which act like other JavaScript primitives, but are composed of other constituent values. This document proposes the first two compound primitive types: `Record` and `Tuple`.
 
 #### Simple primitive types
 
--- a/README.md
+++ b/README.md
@@ -36,8 +36,6 @@ This proposal introduces two new deeply immutable data structures to JavaScript:
 
 Records and Tuples can only contain primitives and other Records and Tuples. You could think of Records and Tuples as "compound primitives". By being thoroughly based on primitives, not objects, Records and Tuples are deeply immutable.
 
-Additionally, this proposal introduces a third new primitive type (`Box`) to ergonomically store object references inside Records and Tuples: `#{ prop: Box(object) }`.
-
 Records and Tuples support comfortable idioms for construction, manipulation and use, similar to working with objects and Arrays. They are compared deeply by their contents, rather than by their identity.
 
 JavaScript engines may perform certain optimizations on construction, manipulation and comparison of Records and Tuples, analogous to the way Strings are often implemented in JS engines. (It should be understood that these optimizations are not guaranteed.)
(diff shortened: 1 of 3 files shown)


tutorial/en.md
```diff
@@ -337,7 +337,7 @@ function getNbVisitByUser(username) {
 function visit(username) {
     // we swap our state with the new one
     appState = #{
-        // We use pushed() to return the new appState.userByVisit record
+        // We spread-append to return the new appState.userByVisit record
         // after pushing a new element to it
         userByVisit: appState.userByVisit.pushed(username),
```
nicolo-ribaudo, line 342: This still uses `pushed`

tutorial/en.md
```diff
@@ -522,114 +520,56 @@ let appState = #{
 };
 ```
 
-What we can do is modify our example so that the mutable parts are wrapped in boxes:
+What we can do is modify our example so that the mutable parts are referenced with a WeakMap:
 
 ```js
+// temporary: until WeakMaps can contain symbols as keys, we need to swap them for Maps
+globalThis.WeakMap = globalThis.Map;
```
nicolo-ribaudo, line 527: This screams "big memory leaks for whoever copies these examples"; can we just use `new Map` for now and move the comment that it's temporary to the `RefsRegistry` definition line?

nicolo-ribaudo:
Could you also update the README as part of this PR? :pray:

rricard:
@nicolo-ribaudo done

README.md
```diff
@@ -509,22 +450,19 @@ The [appendix](./NS-Proto-Appendix.md#tuple-prototype) contains a full descripti
...
 ```js
 assert(typeof #{ a: 1 } === "record");
 assert(typeof #[1, 2]   === "tuple");
-assert(typeof Box({}) === "box");
 ```
 
 ## Usage in {`Map`|`Set`|`WeakMap`|`WeakSet`}
 
-It is possible to use a `Record`, `Tuple` or `Box` as a key in a `Map`, and as a value in a `Set`. When using a `Record` or `Tuple` here, they are compared by value.
+It is possible to use a `Record`, `Tuple` as a key in a `Map`, and as a value in a `Set`. When using a `Record` or `Tuple` here, they are compared by value.
```
acutmore, line 462:
```suggestion
It is possible to use a `Record` or `Tuple` as a key in a `Map`, and as a value in a `Set`. When using a `Record` or `Tuple` here, they are compared by value.
```

README.md
```diff
@@ -865,13 +803,10 @@ A new, deeply immutable, compound primitive type data structure, proposed in thi
...
 A new, deeply immutable, compound primitive type data structure, proposed in this document, that is analogous to Array. `#[1, 2, 3, 4]`
 
-#### Box
-
-A new primitive type that can store a reference to an object, proposed in this document. `Box({})`.
 
 #### Compound primitive types
 
-Values which act like other JavaScript primitives, but are composed of other constituent values. This document proposes the first three compound primitive types: `Record`, `Tuple` and `Box`.
+Values which act like other JavaScript primitives, but are composed of other constituent values. This document proposes the first three compound primitive types: `Record` and `Tuple`.
```
acutmore, line 809:
```suggestion
Values which act like other JavaScript primitives, but are composed of other constituent values. This document proposes the first two compound primitive types: `Record` and `Tuple`.
```

PR #279, opened 2021-12-27
Fix Object.keys(#{ a: 1, b: 2 }) === #["a", "b"]

ascoders:
> assert(Object.keys(#{ a: 1, b: 2 }) === #["a", "b"]);

should be

> assert(Object.keys(#{ a: 1, b: 2 }) !== #["a", "b"]);

### DIFF
--- a/README.md
+++ b/README.md
@@ -411,7 +411,7 @@ For integrity, out-of-bounds numerical indexing on Tuples returns `undefined`, r
 `Tuple` values have functionality broadly analogous to `Array`. Similarly, `Record` values are supported by different `Object` static methods.
 
 ```js
-assert(Object.keys(#{ a: 1, b: 2 }) !== #["a", "b"]);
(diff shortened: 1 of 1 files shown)


acutmore:
An alternative fix could be:

```js
assert(#[…Object.keys(#{ a: 1, b: 2 })] === #["a", "b"]);
```

nicolo-ribaudo:
I'd prefer `assert.deepEqual(Object.keys(#{ a: 1, b: 2 }), ["a", "b"])`.

ascoders:
> I'd prefer `assert.deepEqual(Object.keys(#{ a: 1, b: 2 }), ["a", "b"])`.

Yes i agree, the key point here is to explain that `Record` can be used by `Object.keys` that treated as ordinary objects. And this pr updated just now.

nicolo-ribaudo (approved):

PR #285, opened 2022-01-03
Record exotic object [[Delete]] was using `?` for a `!` case

bathos:
AFAICT, RecordHasProperty cannot throw and other places where it gets invoked already use `!`.


nicolo-ribaudo (approved):

nicolo-ribaudo:
Thanks!

PR #294, opened 2022-03-17
Fixed two typos in cookbook/en.md

ioannad:
Some initial typos found while turning the examples into implementation tests.


nicolo-ribaudo (approved):
Thanks!

PR #296, opened 2022-04-05
Update README.md

unadlib:
Fix the Record example typo.


rickbutton:
`Record.from` does not exist, the example is correct in this case.

ljharb:
@rickbutton if it doesn't exist, then the TypeError isn't for that reason in the example

rickbutton:
oh, I read the diff in the opposite direction.

rickbutton (approved):

PR #303, opened 2022-05-12
spec: make sure we get string keys in Record.fromEntries

rricard:
This coerces keys to strings like Object.fromEntries would do while rejecting symbols as they cannot be record keys

### DIFF
--- a/spec/immutable-data-structures.html
+++ b/spec/immutable-data-structures.html
@@ -49,9 +49,8 @@
           1. Perform ? RequireObjectCoercible(_iterable_).
           1. Let _fields_ be a new empty List.
           1. Let _adder_ be a new Abstract Closure with parameters (_key_, _value_) that captures _fields_ and performs the following steps when called:
-            1. If Type(_key_) is Symbol, throw a *TypeError* exception.
             1. If Type(_value_) is Object, throw a *TypeError* exception.
-            1. Let _keyString_ be ? ToPropertyKey(_key_)
+            1. Let _keyString_ be ? ToString(_key_).
             1. Let _field_ be { [[Key]]: _keyString_, [[Value]]: _value_ }.
             1. Append _field_ to the end of list _fields_.
           1. Perform ! AddEntriesFromIterable(*undefined*, _iterable_, _adder_).
--- a/spec/immutable-data-structures.html
+++ b/spec/immutable-data-structures.html
@@ -53,7 +53,7 @@
             1. If Type(_value_) is Object, throw a *TypeError* exception.
             1. Let _field_ be { [[Key]]: _keyString_, [[Value]]: _value_ }.
             1. Append _field_ to the end of list _fields_.
-          1. Perform ! AddEntriesFromIterable(*undefined*, _iterable_, _adder_).
+          1. Perform ? AddEntriesFromIterable(*undefined*, _iterable_, _adder_).
           1. Return ! CreateRecord(_fields_).
         </emu-alg>
         <emu-note>
--- a/spec/immutable-data-structures.html
+++ b/spec/immutable-data-structures.html
@@ -49,8 +49,8 @@
           1. Perform ? RequireObjectCoercible(_iterable_).
           1. Let _fields_ be a new empty List.
           1. Let _adder_ be a new Abstract Closure with parameters (_key_, _value_) that captures _fields_ and performs the following steps when called:
-            1. If Type(_value_) is Object, throw a *TypeError* exception.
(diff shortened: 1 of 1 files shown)


spec/immutable-data-structures.html
```diff
@@ -49,8 +49,10 @@ <h1>Record.fromEntries ( _iterable_ )</h1>
           1. Perform ? RequireObjectCoercible(_iterable_).
           1. Let _fields_ be a new empty List.
           1. Let _adder_ be a new Abstract Closure with parameters (_key_, _value_) that captures _fields_ and performs the following steps when called:
+            1. If Type(_key_) is Symbol, throw a *TypeError* exception.
             1. If Type(_value_) is Object, throw a *TypeError* exception.
-            1. Let _field_ be { [[Key]]: _key_, [[Value]]: _value_ }.
+            1. Let _keyString_ be ? ToString(_key_)
```
ljharb, line 54: Doesn’t this already handle the Symbol case?

spec/immutable-data-structures.html
```diff
@@ -49,8 +49,10 @@ <h1>Record.fromEntries ( _iterable_ )</h1>
           1. Perform ? RequireObjectCoercible(_iterable_).
           1. Let _fields_ be a new empty List.
           1. Let _adder_ be a new Abstract Closure with parameters (_key_, _value_) that captures _fields_ and performs the following steps when called:
+            1. If Type(_key_) is Symbol, throw a *TypeError* exception.
```
ljharb, line 52:
```suggestion
```

acutmore, line 52: we could drop this as the `toString` AO we are using does this.

Also `Object.fromEntries` does https://tc39.es/ecma262/#sec-topropertykey calling `toPrimitive`

spec/immutable-data-structures.html
rricard: yes, but we figured out that for symmetry with Object.fromEntries, we should use ToPropertyKey instead and then the assertion for a symbol becomes needed again

acutmore (approved):

spec/immutable-data-structures.html
```diff
@@ -49,8 +49,10 @@ <h1>Record.fromEntries ( _iterable_ )</h1>
...
             1. If Type(_value_) is Object, throw a *TypeError* exception.
-            1. Let _field_ be { [[Key]]: _key_, [[Value]]: _value_ }.
+            1. Let _keyString_ be ? ToPropertyKey(_key_)
```
ljharb, line 54: How can ToPropertyKey throw if you know key isn’t an object?

ljharb: Then this call could produce a Symbol, so the symbol check must be done after this line.

acutmore: great spot!

acutmore: I've gone back to `ToString` this time without the extra superfluous check for symbol.

I am now also wondering if we should do the key processing before throwing if the value is not an object. As the key is _earlier_ in the array.

ljharb: Yes, 1000%

acutmore: done!

spec/immutable-data-structures.html
```diff
@@ -49,8 +49,9 @@ <h1>Record.fromEntries ( _iterable_ )</h1>
           1. Perform ? RequireObjectCoercible(_iterable_).
           1. Let _fields_ be a new empty List.
           1. Let _adder_ be a new Abstract Closure with parameters (_key_, _value_) that captures _fields_ and performs the following steps when called:
+            1. Let _keyString_ be ? ToString(_key_).
             1. If Type(_value_) is Object, throw a *TypeError* exception.
-            1. Let _field_ be { [[Key]]: _key_, [[Value]]: _value_ }.
+            1. Let _field_ be { [[Key]]: _keyString_, [[Value]]: _value_ }.
             1. Append _field_ to the end of list _fields_.
           1. Perform ! AddEntriesFromIterable(*undefined*, _iterable_, _adder_).
```
ljharb, line 56: I believe this should be a ? now

acutmore: ooo yes - thanks @ljharb

rricard (commented):
thanks for updating it ashley, will need one more champion review now that we both are authors, paging @rickbutton

ljharb (approved):

rickbutton (approved):

PR #321, opened 2022-07-11
no longer throw for toPrimitive of a record wrapper

### LINKED ISSUE #320: ToPrimitive of Record wrapper throws
[ToPrimitive](https://tc39.es/ecma262/#sec-toprimitive) of an object relies on looking up methods on the value, there is no default behavior:

```js
delete Symbol.prototpe.toString;
delete Symbol.prototpe.valueOf;
delete Symbol.prototpe[Symbol.toPrimtive];
Object.setPrototypeOf(Symbol.prototype, null);

Symbol() == Object(Symbol()); // throws
```

This behavior currently happens by default for Record wrappers because they have a null prototype.

```js
#{} == Object(#{}) // throws
```

It seems more future-proof to allow Record wrappers to safely pass `ToPrimitive` by unwrapping to their record primitive content. `ToString` and `ToNumber` can still throw.

We could add support for this either by adding a direct case for handing records to `ToPrimitive`, or always add a non-enumerable `Symbol.toPrimitive` to Record wrappers - though this would add a new hidden intrinsic.

acutmore:
Closes #320 

An alternative approach would be to add `Symbol.toPrimitive` on all Record wrappers, though this would introduce a new hidden intrinsic.


ljharb:
ah, because there’s no Record.prototype to find it on :-/ that definitely won’t gain consensus

PR #325, opened 2022-07-19
various: Spec fixups

rricard:
Fixes found during a spec review with @acutmore

### DIFF
--- a/spec/data-types-and-values.html
+++ b/spec/data-types-and-values.html
@@ -7,7 +7,7 @@
 
     <emu-clause id="sec-ecmascript-language-types-record-type">
       <h1>The Record Type</h1>
-      <p>The Record type is the set of all mappings from Strings to ECMAScript primitive values including Record and Tuple. Each record value holds an associated [[Fields]] List value which is a list of pairs of the form { [[Key]], [[Value]] } where the [[Key]] is a String and [[Value]] is any primitive value. Entries of [[Fields]] are sorted by [[Key]] in code unit order. The [[Fields]] List and its entries are never modified.</p>
+      <p>The Record type is the set of all mappings from Strings to ECMAScript primitive values including Record and Tuple. Each record value holds an associated [[Fields]] List value which is a list of pairs of the form { [[Key]], [[Value]] } where the [[Key]] is a String and [[Value]] is any primitive value. Entries of [[Fields]] are sorted by [[Key]] in ascending code unit order. The [[Fields]] List and its entries are never modified.</p>
 
       <emu-clause id="sec-immutable-types-record-tostring" type="abstract operation">
         <h1>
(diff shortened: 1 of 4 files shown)


spec/data-types-and-values.html
```diff
@@ -7,7 +7,7 @@ <h1>ECMAScript Language Types</h1>
 
     <emu-clause id="sec-ecmascript-language-types-record-type">
       <h1>The Record Type</h1>
-      <p>The Record type is the set of all finite mappings from Strings to ECMAScript primitive values including Record and Tuple. Each record value holds an associated [[Fields]] List value which is a list of pairs of the form { [[Key]], [[Value]] } where the [[Key]] is a String and [[Value]] is any primitive value. Entries of [[Fields]] are sorted by [[Key]] in code unit order. The [[Fields]] List and its entries are never modified.</p>
+      <p>The Record type is the set of all mappings from Strings to ECMAScript primitive values including Record and Tuple. Each record value holds an associated [[Fields]] List value which is a list of pairs of the form { [[Key]], [[Value]] } where the [[Key]] is a String and [[Value]] is any primitive value. Entries of [[Fields]] are sorted by [[Key]] in ascending code unit order. The [[Fields]] List and its entries are never modified.</p>
```
nicolo-ribaudo, line 10: What's the reason for this change? Records cannot have an infinite number of keys, so the old definition was correct.

This "finite" does not refer to the number of records, but to the number of entries in each record.

acutmore: I thought the mappings would be infinite because there are an infinite number of ECMAScript primitive values  that the strings could map to. Though I guess if we take the mapping as each string to the set of primitives, then that is finite.

nicolo-ribaudo: There are infinite mappings because there are infinite primitive values, but each one of them is finite. Technically, Records are all the finite subsets of "Set of Strings"x"Set of ECMAScript language values" such that there are no two identical keys in the same record, and Tuples are all the finite ordered subsets of "Set of ECMAScript language values".

acutmore: To check if I've understood. The following sequence is what I'm imagining:

```
Record:
  # no keys
  {}
  # 1 key with string length = 0
  { ("", EPV) }
  # 1 key with string length = 1
  { ("\u{0}", EPV) }
  { ("\u{1}", EPV) }
  ...
  # 1 key with string length = 2
  { ("\u{0}\u{0}", EPV) }
  { ("\u{0}\u{1}", EPV) }
  ...
  # 2 keys with string length 0 and 1
  { ("", EPV), ("\u{0}", EPV) }
  { ("", EPV), ("\u{1}", EPV) }
  ...
```

Gradually enumerating all "finite subsets of "Set of Strings"x"Set of ECMAScript language values" such that there are no two identical keys in the same record,".

nicolo-ribaudo: Exactly! You can go on for infinite steps, but you will never get an infinite record. It's like every natural number is finite, but there is an infinite number of them.

rricard: Thank you for the math lesson @nicolo-ribaudo, I'll revert that part!

acutmore (approved):

PR #331, opened 2022-07-22
fix ecmarkup syntax & pin biblio

bakkot:
First commit fixes the syntax for Records in header types to match 262 (and to make ecmarkup happy).

Second commit pins the biblio, as is [recommended in its readme](https://www.npmjs.com/package/@tc39/ecma262-biblio).


acutmore (approved):
Thank you @bakkot !!

PR #343, opened 2022-08-09
fix record-exotic-objects-defineownproperty

acutmore:
Fixes #342

### DIFF
--- a/spec/ordinary-and-exotic-object-behaviours.html
+++ b/spec/ordinary-and-exotic-object-behaviours.html
@@ -56,7 +56,7 @@
           1. Let _current_ be ! RecordGet(_R_, _P_).
           1. If _current_ is ~empty~ return *false*.
           1. If _Desc_.[[Value]] is present, return ! SameValue(_Desc_.[[Value]], _current_).
-          1. Return *false*.
+          1. Return *true*.
         </emu-alg>
       </emu-clause>
 


anba:
`[[DefineOwnProperty]]` can probably be simplified to:
```
1. Assert: IsPropertyKey(P) is true.
2. Let current be R.[[GetOwnProperty]](P).
3. Return IsCompatiblePropertyDescriptor ( false, Desc, current ).
```

But please double-check I didn't miss some corner cases.

ljharb:
I'd probably keep the early returns for clarity - something like:
```
1. Assert: IsPropertyKey(_P_) is true. (this line won't be necessary once the AO is using typed headers)
1. If Type(_P_) is not String, return *false*.
1. If IsDataDescriptor(_Desc_) is *false*, return *false*.
1. If RecordHasProperty(_P_, _R_) is *false*, return *false*.
1. Let _current_ be _R_.[[GetOwnProperty]](_P_).
1. Return IsCompatiblePropertyDescriptor (*false*, _Desc_, _current_).
```

anba:
> `1. If IsDataDescriptor(_Desc_) is *false*, return *false*.`

Needs to be:
`IsAccessorDescriptor(_Desc_) is *true*, return *false*.`

to correctly handle generic descriptors.

ljharb:
ah, fair

rricard (approved):

acutmore:
Thanks for the further suggestions @anba and @ljharb , as this fix is currently a one liner. I'll merge this PR for now. And do the refactoring as a separate PR.

PR #347, opened 2022-08-10
[spec]: fix step numbers

ljharb:
(except for IsConcatSpreadable, which should be 23.1.3.2.1, but would need to be moved to its own imported file first)

### DIFF
--- a/spec/abstract-operations.html
+++ b/spec/abstract-operations.html
@@ -850,7 +850,7 @@
     </emu-clause>
   </emu-clause>
 
-  <emu-clause id="sec-operations-on-records">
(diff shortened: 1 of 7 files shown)


acutmore:
@ljharb what do you think we should do for the new top level section for "Immutable Data Structures" https://tc39.es/proposal-record-tuple/#sec-primitive-data-structures

Should it be moved into an existing section, or should new sections always be added at the end to avoid impacting existing spec numbers.

ljharb:
I don't think we need to worry about numbering - it should be inserted wherever it makes the most sense.

acutmore (approved):

PR #353, opened 2022-09-02
[spec] fix tuple-exotic-objects-defineownproperty

### LINKED ISSUE #352: `DefineOwnProperty` for tuples is inconsistent
The following is almost identical to Issue #342, but for tuples:

Step 13 of section 10.4.9.3 in the proposal, ` [[DefineOwnProperty]] ( P, Desc )`, is: "Return false". This case is reached when `P` is an existing property in `R` and `Desc` doesn't have a `[[Value]]` field. 

This isn't consistent with the default behavior for objects; #343 fixed it for record exotic objects but not for tuple exotic objects.

acutmore:
Fixes #352 

An empty property descriptor should return true.

cc: @catamorphism

### DIFF
--- a/spec/ordinary-and-exotic-object-behaviours.html
+++ b/spec/ordinary-and-exotic-object-behaviours.html
@@ -251,8 +251,7 @@
           1. If Type(_P_) is a Symbol, return *false*.
           1. If _P_ is *"length"*, then
             1. Let _lengthDesc_ be ! _T_.[[GetOwnProperty]](_P_).
-            1. Let _extensible_ be _T_.[[Extensible]].
-            1. Return IsCompatiblePropertyDescriptor(_extensible_, _Desc_, _lengthDesc_).
+            1. Return IsCompatiblePropertyDescriptor(*false*, _Desc_, _lengthDesc_).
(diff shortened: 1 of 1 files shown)


spec/ordinary-and-exotic-object-behaviours.html
```diff
@@ -251,8 +251,7 @@ <h1>[[DefineOwnProperty]] ( _P_, _Desc_ )</h1>
           1. If Type(_P_) is a Symbol, return *false*.
           1. If _P_ is *"length"*, then
             1. Let _lengthDesc_ be ! _T_.[[GetOwnProperty]](_P_).
-            1. Let _extensible_ be _T_.[[Extensible]].
-            1. Return IsCompatiblePropertyDescriptor(_extensible_, _Desc_, _lengthDesc_).
+            1. Return IsCompatiblePropertyDescriptor(*false*, _Desc_, _lengthDesc_).
```
acutmore, line 254: unrelated. Tidy up while here, `extensible` is always false, looking it up implies it's dynamic.

rricard (approved):

PR #356, opened 2022-09-06
spec: fix isValidTupleIndex

### LINKED ISSUE #355: IsValidTupleIndex needs to check for non-integral numbers
See [IsValidIntegerIndex](https://tc39.es/ecma262/#sec-isvalidintegerindex), step 2:
> 2. If [IsIntegralNumber](https://tc39.es/ecma262/#sec-isintegralnumber)(index) is false, return false.

The same check must be added to IsValidTupleIndex.

Additionally the description for IsValidTupleIndex and TupleGet needs to be adjusted to declare the `numericIndex` argument as a "Number" instead of an "integral Number".

acutmore:
Fixes #355 

cc: @anba

### DIFF
--- a/spec/ordinary-and-exotic-object-behaviours.html
+++ b/spec/ordinary-and-exotic-object-behaviours.html
@@ -379,14 +379,16 @@
           <ins>
             IsValidTupleIndex (
               _T_: a Tuple object,
-              _numericIndex_: an integral Number,
+              _index_: a Number,
             ): a Boolean
           </ins>
         </h1>
-        <dl class="header"></dl>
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo (approved):

anba:
`TupleGet` also needs to be updated to accept a `Number` instead of `integral Number`.

acutmore:
> `TupleGet` also needs to be updated to accept a `Number` instead of `integral Number`.

thanks for spotting that too! Fixed now.

ljharb (approved):

PR #362, opened 2022-10-26
Add explicit tuple methods steps

### LINKED ISSUE #323: Ensure correct behavior of Tuple methods that refer to the definition of Array methods
Some of the tuple methods, such as [Tuple.prototype.map](https://tc39.es/proposal-record-tuple/#sec-tuple.prototype.map), are defined with explicit steps and, when they have a callback, they pass the result of `thisTupleValue(this)` as the last parameter.
Other tuple methods, such as [Tuple.prototype.forEach](https://tc39.es/proposal-record-tuple/#sec-tuple.prototype.foreach), are defined as a diff from the relative Array method, with references to `this` replaced with `thisTupleVaue(this)`. However, array methods call `ToObject(this)` (which thus ends up being `ToObject(thisTupleVaue(this)))`) and pass the resulting object as the last parameter to the callback function.

In practice, this means that `#[1].forEach((v, i, tup) => console.log(typeof tup))` logs `"object"` instead of `"tuple"`: we should fix how we define those methods to make sure that the last parameter is correctly a primitive tuple.

Ref https://github.com/tc39/proposal-record-tuple/pull/317#discussion_r920270217

nicolo-ribaudo:
Closes #323.

This has normative effect, because tuples are not casted anymore to objects when passing it to the various callbacks.

### DIFF
--- a/spec/immutable-data-structures.html
+++ b/spec/immutable-data-structures.html
@@ -218,8 +218,19 @@
       </emu-clause>
       <emu-clause id="sec-tuple.prototype.at">
         <h1>Tuple.prototype.at ( _index_ )</h1>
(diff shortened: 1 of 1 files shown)


ljharb (approved):
This is a good and necessary change; callback authors should work with tuple primitives instead of boxed tuples.

acutmore (approved):

